PR #8, opened 2016-08-16
Add some plaintext content to the test site

dwijnand:
.. for greppability purposes, for paradox beginners (such as myself).


dwijnand:
Someone enabled Travis :) (and there's no .travis.yml file so it's failing)

PR #9, opened 2016-08-17
Introducing: Leady McBreadcrumbs

dwijnand:
If a documentation site is contained within an enclosing website it's very pleasant if the breadcrumbs allows one to navigate back up.

This introduces this as a first class concept within Paradox.

---

Despite how nicely Paradox is coded, I had a little trouble figuring out how to implement this. Please consider this a first draft and I very much welcome feedback, including feedback about how to completely re-do this. The only thing I'm happy with is that it achieves the end goal (in the test).


pvlugter:
Nice. LGTM.

pvlugter:
I wonder if we could properly support "template extensions" like this. Making it easier to extend, or add new features. Same for connecting sbt settings to directives.

PR #18, opened 2016-08-31
Allow for non-H1 headers

### LINKED ISSUE #17: Allow non-h1 to be auto title
Currently each page requires H1 level header.
Allow the first h2, h3, or h4 to be a title too.

dwijnand:
Fixes #17


eed3si9n:
LGTM

pvlugter:
LGTM2.

I'm trying to remember why it was restricted to H1 — it was obviously a conscious decision — but I can't think what it was for now. I see there's a `TODO: get default label node from page index link?` comment, so maybe that was the original idea.

PR #22, opened 2016-09-13
Aggregate from multiple snip sections

jonas:


plugin/src/sbt-test/paradox/snippets/src/main/paradox/multiple.md
```diff
@@ -1 +1,3 @@
 @@ snip [multiple snippet](../../test/scala/Multiple.scala) { #multiple }
+
+@@ snip [comma separated snippet](../../test/scala/Multiple.scala) { #parseint-imports #parseint-def }
```
eed3si9n: Could you add a test for the case that has `type=conf`?

eed3si9n:
Thanks a lot for adding this feature!

ktoso:
Thanks a lot for helping out on this front @jonas!

plugin/src/sbt-test/paradox/snippets/src/main/paradox/multiple.md
jonas: Done. Let me know if you want me to squash it.

eed3si9n:
LGTM

eed3si9n:
@jonas Do you want me to cut a release out as is, or are you working on more a PR for the other issue?

jonas:
@eed3si9n Two potential things would help a lot converting the Akka HTTP docs to Paradox at this point.

- Cross-references which automatically show the titles being linked to. Right now the `@ref[]` directive requires the link text. It would be nice to support something like `@ref[title-anchor]` but this might be a lot of work since there would have to be some `@anchor[title-anchor]` etc.

- Allow custom start/end delimiters or regexes for snippets. This will help a lot with extracting the [API](http://doc.akka.io/docs/akka/2.4/scala/http/routing-dsl/directives/basic-directives/cancelRejection.html#cancelrejection) for the many directives. This is currently handled by `includecode2` which supports `def` and `val`. Note this might be all moot since I think there's a long-term goal to move more of the directive documentation to java/scala doc AFAIK (see akka/akka-http#11)

The first one can be considered nice-to-have and might even be out of scope for Paradox, however, the latter would remove the need to mess up the API code with lots of markers or extract the snippets via an SBT task before running Paradox.

eed3si9n:
@jonas Could you open new Github issues to discuss them?

jonas:
@eed3si9n Opened related tickets. When you have time please cut a release so this won't block akka/akka-http#1

eed3si9n:
@jonas Paradox 0.2.2 is pushed to Sonatype on its way to Maven Central.
https://oss.sonatype.org/content/repositories/public/com/lightbend/paradox/paradox_2.10/

PR #28, opened 2016-09-26
Properties front matter

### LINKED ISSUE #15: properties front matter
I want to be able to pass in template values at the page level like this:

```
---
layout: something
out: foo.html
---
```

I've implemented that for Pamflet similar to Jekyll's [YAML front matter](https://jekyllrb.com/docs/frontmatter/), and it's pretty useful.

gsechaud:
Global support for properties at page level.
For now, only support for "out" and "layout" properties.

resolve #15


README.markdown
```diff
@@ -48,6 +48,26 @@ $page.title$
...
 is substituted with the title of the page.
 
+### Properties front matter
+
+Paradox allows to specify some properties at page level using `---` delimiters
+
+The `out` property allows to rename the target name of the current file.
+And the `layout` allows to specify the layout we want to be used for this particular page, the layouts are placed by default in the `target/paradox/theme` folder:
+
+```scala
```
jonas: This should be `markdown`?

gsechaud: Yes of course, my bad!

core/src/main/scala/com/lightbend/paradox/template/PageTemplateSystem.scala
```diff
@@ -20,20 +20,33 @@ import java.io.File
...
+  }
+
+  val cache: TrieMap[String, PageTemplate] = TrieMap()
+}
 
 /**
  * Page template writer.
  */
-class PageTemplate(directory: File, startDelimiter: Char = '$', stopDelimiter: Char = '$', name: String = "page") {
+class PageTemplate(directory: File, name: String = PageTemplate.DefaultName, startDelimiter: Char = '$', stopDelimiter: Char = '$') {
```
eed3si9n: Could you add test for layout feature?

eed3si9n: Here's what I used for testing https://github.com/eed3si9n/tech-hub-landing-page/tree/wip/layout_check, which didn't work for me.

eed3si9n: After fixing the typo in my test, it seems to be working now. We should still add tests though.

core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
```diff
@@ -0,0 +1,54 @@
...
+      case _ => Frontin(Map.empty[String, String], str)
+    }
+
+  def loadProperties(str: Option[String]): Map[String, String] = str match {
+    case None => Map.empty[String, String]
+    case _ =>
+      val p = new java.util.Properties
+      for (s <- str) {
+        val q = new java.util.Properties
+        q.load(new StringReader(s))
```
jonas: Given that Typesafe Config is the most awesome config reader on the JVM, it would be interesting to eventually switch to that for managing properties in Paradox and allow more flexible types and loading of properties.

pvlugter: Yes, would be really nice to use Typesafe Config. There could be some inspiration in the way Stefan's Ornate uses Config: https://github.com/szeiger/ornate

eed3si9n:
I'm going to merge as is, and we can add the HOCON (Typesafe Config) support later.

Issue #33, opened 2016-10-12
Provide proper error message when reference cannot be resolved


jrudolph:
Currently, we are getting errors like this:

```
java.util.NoSuchElementException: key not found: ../client-side/index.md#http-client-side-java
	at scala.collection.MapLike$class.default(MapLike.scala:228)
	at scala.collection.AbstractMap.default(Map.scala:58)
	at scala.collection.MapLike$class.apply(MapLike.scala:141)
	at scala.collection.AbstractMap.apply(Map.scala:58)
	at com.lightbend.paradox.markdown.RefDirective.render(Directive.scala:91)
	at com.lightbend.paradox.markdown.DirectiveSerializer.visit(Directive.scala:39)
	at org.pegdown.ToHtmlSerializer.visit(ToHtmlSerializer.java:386)
	at org.pegdown.ast.DirectiveNode.accept(DirectiveNode.java:58)
	at org.pegdown.ToHtmlSerializer.visitChildren(ToHtmlSerializer.java:397)
	at org.pegdown.ToHtmlSerializer.visit(ToHtmlSerializer.java:381)
	at org.pegdown.ast.SuperNode.accept(SuperNode.java:43)
	at org.pegdown.ToHtmlSerializer.visitChildren(ToHtmlSerializer.java:397)
	at org.pegdown.ToHtmlSerializer.printTag(ToHtmlSerializer.java:418)
	at org.pegdown.ToHtmlSerializer.printBreakBeforeTag(ToHtmlSerializer.java:425)
	at org.pegdown.ToHtmlSerializer.visit(ToHtmlSerializer.java:190)
	at org.pegdown.ast.ParaNode.accept(ParaNode.java:35)
	at org.pegdown.ToHtmlSerializer.visitChildren(ToHtmlSerializer.java:397)
	at org.pegdown.ToHtmlSerializer.visit(ToHtmlSerializer.java:89)
	at org.pegdown.ast.RootNode.accept(RootNode.java:51)
	at org.pegdown.ToHtmlSerializer.toHtml(ToHtmlSerializer.java:70)
	at com.lightbend.paradox.markdown.Writer.write(Writer.scala:43)
	at com.lightbend.paradox.ParadoxProcessor$PageContents.<init>(ParadoxProcessor.scala:80)
	at com.lightbend.paradox.ParadoxProcessor.com$lightbend$paradox$ParadoxProcessor$$render$1(ParadoxProcessor.scala:60)
[...]

PR #34, opened 2016-10-12
Fix ref bug with anchor

gsechaud:
Resolve https://github.com/lightbend/paradox/issues/33


jrudolph:
Cool, @gsechaud. That seems to work for akka-http (though, there seems to be another problem that files from the `lib` directory are not copied to the target any more).

gsechaud:
Yes this is a bug introduced by #28. This should be resolved by #30 for the next version of paradox.

jrudolph:
Thanks for the explanation.

PR #39, opened 2016-10-24
Misc fixes, improvements and extensions

sirthias:
I'd like to propose a few changes/extensions that came out of my work on the [swave](http://swave.io) docs so far.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -234,11 +234,15 @@ case class GitHubDirective(currentPath: String, variables: Map[String, String]) 
  *
  * Extracts snippets from source files into verbatim blocks.
  */
-case class SnipDirective(page: Page) extends LeafBlockDirective("snip") {
+case class SnipDirective(page: Page, variables: Map[String, String]) extends LeafBlockDirective("snip") {
+  private lazy val snippetBase =
+    new File(variables.getOrElse("snippet.base_dir", sys.error("Property `snippet.base_dir` is not defined")))
   def render(node: DirectiveNode, visitor: Visitor, printer: Printer): Unit = {
     try {
       val labels = node.attributes.values("identifier").asScala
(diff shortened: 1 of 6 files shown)


ktoso:
Good idea with the base_url's, me likes 👍

dwijnand:
Generally LGTM, but this further undermines Paradox's principle of the markdown being usable as source on GitHub, which I think is a shame.

sirthias:
There is a tradeoff here, between keeping the links in the docs DRY and making them github friendly. I like that paradox gives me as a user the ability to choose my own sweet spot on this tradeoff continuum. Depending on how the dependent project chooses to manage itself the github linkability might be really important or merely nice to have.
I like that paradox isn't overly opinionated in this regard.

dwijnand:
I see what you're saying.

I guess I highly value begin GitHub friendly, so I want to keep Paradox opinionated as such.

But if this helps adoption then ok.

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -234,11 +234,15 @@ case class GitHubDirective(currentPath: String, variables: Map[String, String])
  *
  * Extracts snippets from source files into verbatim blocks.
  */
-case class SnipDirective(page: Page) extends LeafBlockDirective("snip") {
+case class SnipDirective(page: Page, variables: Map[String, String]) extends LeafBlockDirective("snip") {
+  private lazy val snippetBase =
+    new File(variables.getOrElse("snippet.base_dir", sys.error("Property `snippet.base_dir` is not defined")))
```
dwijnand: Won't this explode in the cases where you don't define it? (Same for image.base_url)

sirthias: Note: It's a `lazy val` though!

dwijnand: that's it, thanks :)

PR #51, opened 2016-11-03
Consistently render content inside callouts, fixes #49

### LINKED ISSUE #49: Consistently render content inside @@@ directives
As noted here: akka/akka-http@f349d908d114b8d42b1287be662a26b362ef4d8c

>  There's a strange thing with how the Pegdown parser handles content
    inside a container directives (`@@@`) where it will honour the new lines
    if there's no empty line between the directive marker and the first
    text.

It would be nice if Paradox would render
```
@@@ note
A
B
@@@
```
and
```
@@@ note

A
B

@@@
```
in the same way.

jonas:
Wrap inner block content in new lines to force consistent parsing of @@@
directives in particular callouts which renders the content directly.

### DIFF
--- a/core/src/main/java/org/pegdown/ParserWithDirectives.java
+++ b/core/src/main/java/org/pegdown/ParserWithDirectives.java
@@ -154,7 +154,7 @@ public class ParserWithDirectives extends Parser {
             push(getContext().getCurrentIndex()),
             contents.clearContents(),
             OneOrMore(TestNot(ContainerBlockDirectiveEnd(markerLength)), ANY, contents.append(matchedChar())),
-            push(parseBlockContents(markerLength, contents.appended("\n")))
+            push(parseBlockContents(markerLength, contents.appended("\n\n")))
         );
     }
 
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Writer.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Writer.scala
@@ -119,7 +119,7 @@ object Writer {
     VarsDirective(context.properties),
     CalloutDirective("note", "Note"),
     CalloutDirective("warning", "Warning"),
-    WrapDirective("div"), WrapDirective("p"))
+    WrapDirective("div"))
 
   class DefaultLinkRenderer(context: Context) extends LinkRenderer {
     private lazy val imgBase = {
--- a/core/src/test/scala/com/lightbend/paradox/markdown/CalloutDirectiveSpec.scala
+++ b/core/src/test/scala/com/lightbend/paradox/markdown/CalloutDirectiveSpec.scala
@@ -34,7 +34,36 @@ class CalloutDirectiveSpec extends MarkdownBaseSpec {
       """) shouldEqual html("""
       |<div class="callout note">
       |<div class="callout-title">Note</div>
-      |Latest version is 1.2.3
+      |<p>Latest version is 1.2.3</p>
+      |</div>
+      """)
+
+    markdown("""
+      |@@@ note
+      |
+      |Latest version is @var[version]
+      |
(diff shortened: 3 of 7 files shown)


core/src/main/java/org/pegdown/ParserWithDirectives.java
```diff
@@ -162,7 +162,7 @@ public Node parseBlockContents(Var<Integer> markerLength, StringBuilderVar block
         // recursively parse inner contents
         Context<Object> context = getContext();
         Integer length = markerLength.get();
-        String contents = block.getString();
+        String contents = "\n" + block.getString() + "\n";
```
eed3si9n: Wouldn't this have impact on the rendering if we implemented a container block that behaves like "```" like 

```
@@@console
> println("hello")
@@@
```

jonas: Yes, it would have an impact but if you want to do something like tut you have to wrap internal content with "```" anyway to keep the preformatted text, for example the above would pass an HTML blockquote tag to the console directive.

BTW, I had missed that block directives already append a `\n` to the content so the fix can be simplified even further.

Anyway, if this is a showstopper I can submit a PR with tests and documentation instead.

eed3si9n: If the concern is mostly around callout-style markups, would it make sense to implement a utility function that the custom markups can use to opt into blank line leniency?

jonas: I found this behaviour surprising when I used the callouts in Akka HTTP which is why I reported it as a bug and made a PR. Since Paradox already appends `\n\n` to pages and appends `\n` to block content, having it instead append `\n\n` to block content (in the updated patch) would IMO make things more consistent.

Why should content inside block directives render different as if it was put in a standalone page?

The current behaviour is:
```
@@@ note
A

B
@@@
```
renders as
```
<div ...>
<div ...>title</div>
<p>A</A>
B
</div>
```

eed3si9n: @sirthias Do you have an opinion on this topic as we've just merged #58, which I think assumes `@@@div` to render as `<div>content</div>` as your spec asserts, as opposed to `<div><p>content</p></div>`.

eed3si9n:
Currently this is failing on 

```
[info] - should render wrapping `div`s *** FAILED ***
[info]   "<div>[
[info]   <p>Simple sentence here.</p>
[info]   ]</div>
[info]   
[info]   " did not equal "<div>[Simple sentence here.]</div>
[info]   
[info]   " (WrapDirectiveSpec.scala:33)
[info] - should render wrapping `p`s with a custom `id` *** FAILED ***
[info]   "<p id="yeah">[</p>
[info]   <p>Simple sentence here.</p>
[info]   <br>
[info]   <br]>
[info]   
[info]   " did not equal "<p id="yeah">[Simple sentence here.</p]>
[info]   
[info]   " (WrapDirectiveSpec.scala:43)
[info] - should support a custom id and custom CSS classes at the same time *** FAILED ***
[info]   "...h" class="red blue">[
[info]   <p>Simple sentence here.</p>
[info]   ]</div>
[info]   
[info]   " did not equal "...h" class="red blue">[Simple sentence here.]</div>
[info]   
[info]   " (WrapDirectiveSpec.scala:53)
```

jonas:
Should I delete the p directive? Nested paragraphs are not valid and div should serve the purpose of wrapping just fine.

eed3si9n:
@jonas Yea. Let's remove p.

jonas:
OK done. Also updated the release notes.

eed3si9n:
Thanks!

PR #52, opened 2016-11-06
Fix #43: Add @javadoc directive

### LINKED ISSUE #43: Add @javadoc directive in analogy to existing @scaladoc
Linking to javadocs with the @scaladoc directive doesn't work properly as the format of deep-links differs between javadoc and scaladoc sites.

For example with this configuration

```
"scaladoc.org.reactivestreams.base_url" -> "http://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/"
```

this snippet

```
@scaladoc:[Publisher](org.reactivestreams.Publisher)
```

gets rendered as

```
http://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/#org.reactivestreams.Publisher
```

when it should be rendered as

```
http://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/org/reactivestreams/Publisher.html
```

jonas:


jroper:
I find linking directly to the javadoc class HTML to give a terrible user experience, because then they don't have the navigation from the frames that is very often needed.

In Lagom and Play, we always link to `index.html?com/example/SomeClass.html`, this links to the frames version. The one problem with this approach is linking to specific methods doesn't work out of the box, to fix this you need to generate javadocs with a little javascript snippet to scroll to the right place on the frame, we do it like this in Lagom:

https://github.com/lagom/lagom/blob/master/project/Doc.scala#L60

dwijnand:
FYI note that the frames are gone, at least in the official Scala API docs: http://www.scala-lang.org/api/current/scala/Int.html

jroper:
Yep but this is for javadocs, not scaladocs.

jonas:
OK, I will update the PR and rebase to add support for referenced source definitions.

core/src/main/scala/com/lightbend/paradox/markdown/Url.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import java.net.{ URI, URISyntaxException }
+
+/**
+ * Small wrapper around URI to help update individual components.
+ */
+case class Url(base: URI) {
+  def withEndingSlash = base.getPath match {
+    case path if path.endsWith("/index.html") => this
+    case path                                 => copy(path = path + "/")
```
jonas: The main goal here is to avoid links that will automatically redirect, yet allow to spell out with `.../index.html` if so desired.

jonas:
I think I've addressed all comments.

- javadoc links uses `?` to show frames
- Fix scaladoc link to work with Scala 2.12

PR #58, opened 2016-11-07
Add directive for wrapping content in `div` or `p` elements ...

sirthias:
… with custom ids and/or CSS classes.

I've always wanted to have the ability to introduce custom container elements (with custom ids/classes) at arbitrary levels in the HTML node tree. Unfortunately pegdown itself doesn't support this, but paradox's directives make it easy.

Hopefully this addition doesn't clash with any output-agnostic vision you might have for paradox as it definitely increases the coupling to HTML.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -406,7 +406,7 @@ case class VarsDirective(variables: Map[String, String]) extends ContainerBlockD
 /**
  * Callout directive.
  *
- * Replaces property values in verbatim blocks.
+ * Renders call-out divs.
  */
 case class CalloutDirective(name: String, defaultTitle: String) extends ContainerBlockDirective(Array(name): _*) {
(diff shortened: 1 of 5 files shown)


eed3si9n:
Should the span directive provide an alternative for `@span ...`? I am wondering if using `@@@span` would break the block.

sirthias:
Yeah, I think you're right. I'll add a respective test and change `@@@span` to `@span`.

sirthias:
Thinking more about it it seems that this feature really hasn't too much value for `span`s, since their content is usually rather small. Also the current directive syntax doesn't really allow for inner content on `InlineDirective`s.

So I simply removed the `span` support completely. Wrapping is thus only supported for `div` and `p` elements which really is what people (i.e. I :) need.

eed3si9n:
ok. Feature-wise, it looks good. We just need documentation.
I just merged a doc for theming (https://github.com/lightbend/paradox/blob/master/docs/src/main/paradox/features/theming.md) into master, so if you want you can rebase and hang this in that page, or a start a new one.

eed3si9n:
Merged! Thanks for the contribution.

PR #59, opened 2016-11-07
Add support for referenced source definition …

sirthias:
… to all directives with `source` semantics, i.e. `@extref`, `@fiddle`, `@github`, `@ref`, `@scaladoc` and `@snip`.

One core markdown feature is support for defining link targets in two ways: either in parens directly following the link text or through a reference key following the link text in brackets.

So far paradox directives only supported the first alternative.
This patch adds the second.

(I'm assuming that directives which specify their source through reference keys would still be completely github-friendly!)

### DIFF
--- a/core/src/main/java/org/pegdown/ParserWithDirectives.java
+++ b/core/src/main/java/org/pegdown/ParserWithDirectives.java
@@ -70,7 +70,7 @@ public class ParserWithDirectives extends Parser {
             directiveMarker,
             DirectiveName(),
             DirectiveLabel(),
-            MaybeDirectiveSource(),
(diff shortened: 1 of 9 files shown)


eed3si9n:
What's the relationship between this and #58?

sirthias:
No relationship. The two PRs are completely independent.
(I erroneously included the commit for this PR in #58 before, this has been fixed now.)

eed3si9n:
Nice addition!

sirthias:
Thanks, Eugene!

pvlugter:
Nice, @sirthias! This was something I wanted to add originally but never got to.

PR #64, opened 2016-11-21
Provide ability to disable prettify code highlighting

jonas:
Either using:

    ```nocode
    ...
    ```

or for snippets:

    @@ snip [apache.log](apache.log) { type=nocode }

---

As discussed in typesafehub/ssl-config#51 Paradox will eagerly highlight anything inside fenched code block. This PR provides an escape hatch for this behaviour.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/StyledVerbatim.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/StyledVerbatim.scala
@@ -65,5 +65,8 @@ abstract class StyledVerbatimSerializer extends VerbatimSerializer {
  */
 object PrettifyVerbatimSerializer extends StyledVerbatimSerializer {
   override def printPreAttributes(printer: Printer): Unit = printClass(printer, "prettyprint")
-  override def printCodeAttributes(printer: Printer, nodeType: String): Unit = printClass(printer, s"language-$nodeType")
+  override def printCodeAttributes(printer: Printer, nodeType: String): Unit = nodeType match {
+    case "text" | "nocode" => printClass(printer, "nocode")
+    case _                 => printClass(printer, s"language-$nodeType")
+  }
 }
--- a/docs/src/main/paradox/features/snippet-inclusion.md
+++ b/docs/src/main/paradox/features/snippet-inclusion.md
@@ -39,6 +39,14 @@ This will be rendered like this:
 
 ![multi_snip](../img/multi_snip.png)
 
+By default, Paradox uses Prettify to highlight code and will try to detect the
+language of the snippet if it is not explicitly provided. In cases where a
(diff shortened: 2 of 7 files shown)


core/src/main/scala/com/lightbend/paradox/markdown/StyledVerbatim.scala
```diff
@@ -65,5 +65,8 @@ abstract class StyledVerbatimSerializer extends VerbatimSerializer {
  */
 object PrettifyVerbatimSerializer extends StyledVerbatimSerializer {
   override def printPreAttributes(printer: Printer): Unit = printClass(printer, "prettyprint")
-  override def printCodeAttributes(printer: Printer, nodeType: String): Unit = printClass(printer, s"language-$nodeType")
+  override def printCodeAttributes(printer: Printer, nodeType: String): Unit = nodeType match {
+    case "nocode" => printClass(printer, s"nocode")
```
jonas: Prettify uses `nocode`, I am open to suggestions for a better name.

jonas: Prettify will still try to highlight it. `nocode` is the only way to turn off this behaviour.

eed3si9n: Can we pass "nocode" to Prettify when the user types in ` ```text ` in addition to `nocode`?

jonas:
Updated to both map `text` and `nocode`.

The alternative would be to always add the `nocode` class when no node type is defined.

jonas:
Rebased and added notes entry

dwijnand:
This at least is where we can mirror and match GHM, so we should pick
whatever works in native GHM to indicate "plain text, don't highlight". I
expect there are a number of alternatives.

eed3si9n:
@dwijnand I'm guessing that Kramdown / Redcarpet or whatever that implements GFM uses ` ``` ` to indicate plain text without special postfix. We can deal with flipping the default behavior of ` ``` ` in another PR, but having a way to force plain text for now is a useful workaround that doesn't require changing all existing text.

https://help.github.com/articles/creating-and-highlighting-code-blocks/

dwijnand:
Linguist I believe is the tool to categorise languages and I assumed it had
an identifier for plain text. I wasn't implying changing every code snippet.

sirthias:
In addition to the per-snippet solution proposed here I think it'd be great if paradox also offered a way to completely disable the existing strong coupling with google-code-prettify. In my project for example we use `highlight.js` instead, which doesn't require `class="prettyprint"` attributes on `pre` elements.

IMHO paradox should offer an SBT setting that allows users to completely switch off any generation-time highlighting support.

PR #68, opened 2016-11-29
Provide url to Github origin markdown files, resolve

### LINKED ISSUE #57: Provide support for origin source URLs
Lagom (and several other projects) render links back to the original markdown source files in GitHub to facilitate user contributions, see the bottom of this page for an example:

http://www.lagomframework.com/documentation/1.2.x/java/WhatDoesLagomMean.html

Paradox needs to support this such that a theme can render that link.

gsechaud:
Resolve #57 

Setup:
- Set `github.base_dir` to select the github repo.
- Set `github.paradox_dir` to select the directory corresponding to paradox (src/main/paradox and build.sbt)

How:
- To display the source link of the current file: `@source[link]()`
- To display the source link of another file, use relative path: `@source[link](../index.md)`

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -18,6 +18,7 @@ package com.lightbend.paradox.markdown
 
 import com.lightbend.paradox.tree.Tree.Location
 import java.io.{ File, FileNotFoundException }
+import java.nio.file.{ Path => jPath }
 
 import org.pegdown.ast._
 import org.pegdown.ast.DirectiveNode.Format._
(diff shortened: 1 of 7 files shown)


dwijnand:
@gsechaud failed on Windows

jonas:
I would suggest to not solve this by adding a new directive but rather to expose a variable to the theme and let the theme insert a link on each page. This would be much simpler to implement and doesn't require adding "boilerplate" to each documentation page.

gsechaud:
Ok I thought that would be nicer to give the possibility to the user to include "whatever" source he wants inside his markdown files, but as you explained, this makes things more complicated and not necessarily useful.

I'll change then asap the structure in order to only deal with source on templates side with themes, not on markdown side.

jonas:
@gsechaud To include a source link we could extend the `@github` directive to support paths. I didn't do that initially but it would be nice to have. Based on the discussion in #60 then an explicit source link could be made using:

```
Link to @github[page source]($page.path$)
```

Issue #72, opened 2016-12-05
Documentation about snippets are wrong


drewhk:
(version 0.2.7)
The docs here says (http://developer.lightbend.com/docs/paradox/latest/features/snippet-inclusion.html):

```
sbt
:   @@snip [build.sbt](/../../../build.sbt) { #setup_example }
Maven
:   @@snip [pom.xml](../../../pom.xml) { #setup_example }
```

But in reality you need a newline:

```
sbt
:   @@snip [build.sbt](/../../../build.sbt) { #setup_example }

Maven
:   @@snip [pom.xml](../../../pom.xml) { #setup_example }
```

2m:
Fixed in #75

2m:
This should be closed and assigned to the next milestone.

dwijnand:
Shipped in v0.2.8

PR #75, opened 2016-12-06
#72 fix docs on snippets

ktoso:
Is this correct @drewhk?

#72 fix docs on snippets


drewhk:
Yes, this is how it works.

2m:
I wonder if you could also replace the picture below the edit to the rendered output.

PR #78, opened 2016-12-08
Code block highligting and snippet doc improvements

jonas:
This updates #75 to fix rendering of the snippet example and no longer use an image for the `@@snip` documentation.

Also changes the default of how fenced code blocks are treated. To see it in action take a look at https://github.com/jonas/ssl-config/commit/f67336a44982dfbfdf80cbd3ab366611da85ed86 and https://jonas.github.io/ssl-config/WSQuickStart.html


dwijnand:
Such a default change should be mentioned in the releases notes. I'm doing that now to release 0.2.8.

PR #82, opened 2016-12-20
fix extref docs

xuwei-k:
- s/scaladoc/extref https://github.com/lightbend/paradox/blob/3546b7534/plugin/src/sbt-test/paradox/parameterized-links/build.sbt#L10
- use "text" syntax highlight. "//" is not comment in this case

## before

before

## after

after

### DIFF
--- a/docs/src/main/paradox/features/linking.md
+++ b/docs/src/main/paradox/features/linking.md
@@ -109,7 +109,9 @@ URL templates can be configured via `extref.<scheme>.base_url` and the
 template may contain one `%s` which is replaced with the scheme specific
 part of the link URL. For example, given the property:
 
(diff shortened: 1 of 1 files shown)


eed3si9n:
The styling has been fixed in https://github.com/lightbend/paradox/pull/78

eed3si9n:
Looks like it got fixed by https://github.com/lightbend/paradox/pull/113:

http://developer.lightbend.com/docs/paradox/latest/features/linking.html

linking-paradox

pvlugter:
The other fix in this PR was the correction from `scaladoc` to `extref`.

eed3si9n:
Re-opened

PR #83, opened 2017-01-06
Multi-Configuration support + Doc overlays

### LINKED ISSUE #54: Documentation overlays
Often we have to provide two sets of documentation, one for Java users, and one for Scala users.  There are some pages that end up being shared between them.  Attempts to include these in a single manual result in weird things like in Akka, it's very easy when navigating through the Java documentation to end up at the Scala documentation because you clicked a link to a shared page, and then clicked the next link which took you to something Scala specific.

Paradox should provide something that allows keeping the rendered manuals entirely separate, but allows sharing pages describing concepts, installation instructions, etc, between the two manuals.  My suggestion would be to allow having a common directory, and then to "overlay" that directory with the language specific documentation.  So for example:

```
manual
  - common
    - index.md
    - gettingstarted
      - installation.md
      - getting-started.md
  - java
    - actions
      - actions.md
      - routing.md
  - scala
    - actions
      - actions.md
      - routing.md
```

If done right, this can actually be very useful beyond just providing multiple programming languages to share resources, it can also be used for providing translations, as the translations can simply overlay the original, allowing things like code snippets from the original to be reused in the translations.  We have successfully applied this approach in the Play documentation.

gsechaud:
Resolve #54


plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxKeys.scala
```diff
@@ -32,6 +32,7 @@ trait ParadoxKeys {
   val paradoxTargetSuffix = settingKey[String]("Target file suffix for HTML files [default = \".html\"].")
   val paradoxTheme = settingKey[Option[ModuleID]]("Web module name of the paradox theme, otherwise local template.")
   val paradoxThemeDirectory = taskKey[File]("Sync combined theme and local template to a directory.")
+  val paradoxOverlayDirectory = settingKey[File]("Directory containing common source files for configuration.")
```
eed3si9n: It should support multiple directories.

plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala
```diff
@@ -71,7 +71,7 @@ object ParadoxPlugin extends AutoPlugin {
     paradoxProperties ++= linkProperties(scalaVersion.value, apiURL.value, scmInfo.value),
 
     sourceDirectory in paradox := sourceDirectory.value / "paradox",
-    sourceDirectories in paradox := Seq((sourceDirectory in paradox).value),
+    sourceDirectories in paradox := Seq(Option((sourceDirectory in paradox).value), paradoxOverlayDirectory.?.value).flatten,
```
eed3si9n: Is "sourceDirectory in paradox" used as the root directory?

gsechaud: Yes "sourceDirectory in paradox" is the main directory of the project. We just add the "paradoxOverlayDirectory" to the sourceDirectories for the current project.

plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala
```diff
@@ -64,7 +71,7 @@ object ParadoxPlugin extends AutoPlugin {
     paradoxProperties ++= linkProperties(scalaVersion.value, apiURL.value, scmInfo.value),
 
     sourceDirectory in paradox := sourceDirectory.value / "paradox",
-    sourceDirectories in paradox := Seq((sourceDirectory in paradox).value),
+    sourceDirectories in paradox := Seq((sourceDirectory in paradox).value) ++ paradoxOverlayDirectories.?.value.getOrElse(Nil),
```
eed3si9n: Instead of doing `++ paradoxOverlayDirectories.?.value.getOrElse(Nil)`, `paradoxOverlayDirectories` should be initialized to `Nil`.  Otherwise the following blows up:

```scala
paradoxOverlayDirectories in DocsFirst += baseDirectory.value / "src" / "commonFirst"
```

docs/src/main/paradox/features/documentation-overlay.md
```diff
@@ -0,0 +1,23 @@
+Documentation overlay
```
eed3si9n: The header can remain "Documentation overlay", but I think the file should be renamed to something shorter like `overlay.md`, or just be merged with the multi configuration page.

PR #85, opened 2017-01-13
Deprecate `@var[foo]` directive in favor of `$foo$`

sirthias:
(This PR builds on https://github.com/lightbend/paradox/pull/60, so please merge that one first!)

As discussed in https://github.com/lightbend/paradox/pull/60 this PR completes the unification of property substitution towards the StringTemplate `$foo$` notation.

In order to not break existing documentation the old way of using `@var[foo]` instead of `$foo$` still works but is not documented anymore.


eed3si9n:
Looking good. Thanks for working on this.

PR #86, opened 2017-01-13
Fix StringTemplate indent bug for snippets

### LINKED ISSUE #84: Snips included with @@snip introduce extra indentation
Seems this problem was introduced with 0.2.8.

screenshot from 2017-01-10 16 27 38

gsechaud:
Resolve #84 

Use the "NoIndentWriter" as a writer for each template

No matter how the theme is built (indentations or not), the generated html pages will be built in order to remove them.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/template/PageTemplateSystem.scala
+++ b/core/src/main/scala/com/lightbend/paradox/template/PageTemplateSystem.scala
@@ -16,10 +16,10 @@
 
 package com.lightbend.paradox.template
 
-import java.io.File
+import java.io.{ File, OutputStreamWriter, FileOutputStream }
 import java.util.{ Map => JMap }
 import org.stringtemplate.v4.misc.STMessage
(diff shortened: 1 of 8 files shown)


dwijnand:
@jrudolph/@ktoso would you be happy with the rendered html not having any indentation at all?

dwijnand:
@gsechaud looks like there's something up with `paradox / site` on windows.

gsechaud:
retest this please

gsechaud:
retest this please

gsechaud:
For comparison:

**Before**:

minimal-example-usualwriter-html

gave the following unexpected generated page due to the indentation:

minimal-example-usualwriter

**With this fix**:

minimal-example-noindentwriter-html

gives the following expected result:

minimal-example-noindentwriter


The difference resides in the "_generation_" of the html pages, not in the creation of the templates/themes. Then no matter how the template is built (with or without indentation), the generated html pages won't include the indentation if there were.

Again here is a comparison for some generated example code before and after to better understand the principle:

generated-usualwriter

generate-noindentwriter

PR #95, opened 2017-02-01
Tab-group switching for snippets

### LINKED ISSUE #11: Make it possible to switch all tabs on a page
As a Java developer, I would find it very annoying to flip all tabs one-by-one to Java for example.

### LINKED ISSUE #63: Tab selection preference should be remembered in a cookie
Depends on https://github.com/lightbend/paradox/issues/11 (about switching all tabs)

If I'm a Java or Scala person, I always want to go to a page and keep seeing the Java (or Scala) snippets, so I don't get angry about clicking those tabs all the time ;-)

Very important feature to not make users angry about the tabs IMO.

gsechaud:
Resolve #11 
Resolve #63 

Note: update paradox version for paradox in the next release to include the working example in documentation.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -311,7 +311,8 @@ case class SnipDirective(page: Page, variables: Map[String, String])
         } else new File(page.file.getParentFile, source)
       val text = Snippet(file, labels)
       val lang = Option(node.attributes.value("type")).getOrElse(Snippet.language(file))
-      new VerbatimNode(text, lang).accept(visitor)
+      val group = Option(node.attributes.value("group")).getOrElse("")
+      new VerbatimGroupNode(text, lang, group).accept(visitor)
     } catch {
       case e: FileNotFoundException =>
         throw new SnipDirective.LinkException(s"Unknown snippet [${e.getMessage}] referenced from [${page.path}]")
--- a/docs/src/main/paradox/features/snippet-inclusion.md
+++ b/docs/src/main/paradox/features/snippet-inclusion.md
@@ -58,6 +58,45 @@ should not be highlighted set `type=text` in the directive's attribute section:
 @@snip [example.log](example.log) { #example-log type=text }
 ```
 
+#### tab switching
+
+It is possible to associate multiple snippets under the same "tag". If some tab of a snippet is switched by the user, all tabs associated with the selected one will be switched as well. To associate snippet tabs under some tag, set the `group` field of the snippet:
+
+@@@vars
+```markdown
+First-java
+:   @@snip [example-first.java](../../resources/tab-switching/examples.java) { #java_first group=java }
+$empty$
+First-scala
+:   @@snip [example-first.scala](../../resources/tab-switching/examples.scala) { #scala_first group=scala }
+$empty$
+Some separator.
+$empty$
+Second-java
+:   @@snip [example-second.java](../../resources/tab-switching/examples.java) { #java_second group=java }
+$empty$
+Second-scala
+:   @@snip [example-second.scala](../../resources/tab-switching/examples.scala) { #scala_second group=scala }
+```
+@@@
+
+The result will be rendered like this (try to switch tabs):
+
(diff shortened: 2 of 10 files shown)


core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -311,7 +311,8 @@ case class SnipDirective(page: Page, variables: Map[String, String])
         } else new File(page.file.getParentFile, source)
       val text = Snippet(file, labels)
       val lang = Option(node.attributes.value("type")).getOrElse(Snippet.language(file))
-      new VerbatimNode(text, lang).accept(visitor)
+      val group = Option(node.attributes.value("group")).getOrElse("")
```
2m: Here we should use the tab name for a group if the group is not explicitly defined. This would allow easier feature use for the documentation writers making docs less verbose.

docs/src/main/paradox/features/snippet-inclusion.md
```diff
@@ -58,6 +58,45 @@ should not be highlighted set `type=text` in the directive's attribute section:
...
+First-scala
+:   @@snip [example-first.scala](../../resources/tab-switching/examples.scala) { #scala_first group=scala }
+
+Some separator.
+
+Second-java
+:   @@snip [example-second.java](../../resources/tab-switching/examples.java) { #java_second group=java }
+
+Second-scala
+:   @@snip [example-second.scala](../../resources/tab-switching/examples.scala) { #scala_second group=scala }
```
2m: AFAIC there is not way to provide grouping for tabs which contain inline text (not a snippet).

2m:
This looks like a great start. One more thing that would be awesome, if we could add support for `@javadoc[Actor](akka.actor.Actor)` and `@scaladoc[Actor](akka.actor.Actor)` having implicit group names of `java` and `scala`. That would allow switching the links to the API docs. If that is too much for this PR, I can create a separate ticket.

plugin/src/sbt-test/paradox/snippets/src/main/paradox/group.md
```diff
@@ -0,0 +1,3 @@
+@@ snip [Group A](../../test/scala/Multiple.scala) { #multiple-something-else group=a }
+
+@@ snip [Group B](../../test/scala/Multiple.scala) { #parseint-imports group=b }
```
2m: This tests if the generated HTML is correct, but will the group=b snippet be hidden by default? I have tried it on a project and it looks like that both snippets are always visible.

Also if there are no tabbed snippets in the page, then a user will have no way of switching between the two. I think we should add a hover box or other interactive control that would allow switching between the alternatives.

docs/src/main/paradox/features/snippet-inclusion.md
pvlugter: Yes, would be good to have an approach that supports anything in the tab, rather than relying on the group attribute in snippets.

pvlugter: Does the current tab grouping allow different types of groups? For example, Java/Scala and sbt/Maven/Gradle on the same page.

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
raboof: The tab name doesn't seem too easy to reach at this point, but defaulting to the `lang` seems reasonable, too - applied to #107

docs/src/main/paradox/features/snippet-inclusion.md
raboof: Indeed tabs with inline text are still not supported, but switching arbitrary text is now available in #107.

Different types of groups on the same page should be no problem.

plugin/src/sbt-test/paradox/snippets/src/main/paradox/group.md
raboof: Added a selectbox for switching in #107 

Actually selecting tabs/alternatives is done in javascript, so isn't easy to incorporate in a test...

2m:
Superseded by #107.

Issue #106, opened 2017-04-18
Text switching


raboof:
Beyond switching tab groups (good to see #95!), it would be interesting to allow switching some of the 'regular text' along with it: the Akka and akka-http documentation refers to types a lot, and these can look different or even _be_ completely different between the Java API and the Scala API.

We'd have to come up with some (hopefully reasonably lightweight) syntax - I'm not aware of any conventions to add this kind of markup to a Markdown document?

pvlugter:
Fixed in #107

Issue #14, opened 2016-08-23
Need PDF output


drewhk:
Many of our users/readers perfer to have a PDF version. We should have some options for this, maybe as simple as emitting LaTeX from the DOM.

eed3si9n:
For Pamflet I've implemented PDF generation by creating a combined markdown, and then passing it to Pandoc. Looks like this - http://www.scala-sbt.org/0.13/docs/sbt-reference.pdf

ktoso:
That would be good enough I think, looks pretty good.
In our community people keep praising the availability of PDF because they
often read the docs like a book.

-- 
Konrad Malawski

From: eugene yokota <notifications@github.com> <notifications@github.com>
Reply: lightbend/paradox
<reply@reply.github.com>
<reply@reply.github.com>
Date: 23 August 2016 at 18:46:47
To: lightbend/paradox <paradox@noreply.github.com>
<paradox@noreply.github.com>
Subject:  Re: [lightbend/paradox] Need PDF output (#14)

For Pamflet I've implemented PDF generation by creating a combined

eed3si9n:
Yea. I think being able to download and carry docs around is a benefit of static generation in general. There are probably a few tweaks we want to allow in Paradox if we want to make combined markdown a thing, like not insisting on having h1 at the top of the page.

raboof:
When groups are used ideally we'd be able to generate a PDF for each group

dwijnand:
> When groups are used ideally we'd be able to generate a PDF for each group

groups?

raboof:
http://developer.lightbend.com/docs/paradox/latest/features/groups.html :)

raboof:
There are a number of ways to approach this, each with their pro's and cons. One is to take the generated HTML and post-process it. If anyone wants to give that a try, https://github.com/raboof/paradox-singlepage could be a starting point (though very incomplete and currently written towards processing the Akka docs)

PR #107, opened 2017-04-19
Switching tabs as well as other fragments

raboof:
Expanding on the proposal in #95, added support for having other parts of
the documentation change as well when switching groups.


raboof:
Refs #106

pvlugter:
The text switching on the spans looks good. Having a general purpose span directive is useful. My first thought for frequently used text groups is whether it would be useful to have it add extra named span directives, rather than just using span classes. Say specifying the possible groups in the build, since these are normally something like Java/Scala and used across the docs, and then having a named span directive like @scala[Scala only text]. So that it will warn you if use an unintended group or have a typo, and the syntax is more direct. Maybe using the group name directly doesn't work and it needs something else, like @for-scala or @scala-only or some other convention. Knowing what the possible groups are at the highest level, to add overall switches at the page level, could also be useful.

raboof:
I like the idea of configuring 'supergroups' in the build (instead of the directive I added for that) and generating directives for them. Using the group name as directive name should be fine (you'll just have to make sure not to choose group names that clash with other directives) - otherwise `@scala-group[Scala only snippet]` might work.

This will also help when adding overall switches, too, indeed - and perhaps when implementing #14 we'll actually need it for generating a PDF for each possible selection.

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -311,7 +311,8 @@ case class SnipDirective(page: Page, variables: Map[String, String])
         } else new File(page.file.getParentFile, source)
       val text = Snippet(file, labels)
       val lang = Option(node.attributes.value("type")).getOrElse(Snippet.language(file))
-      new VerbatimNode(text, lang).accept(visitor)
+      val group = Option(node.attributes.value("group")).getOrElse("")
```
2m: What do you think about the idea mentioned here: https://github.com/lightbend/paradox/pull/95#discussion_r100283649

That would allow easier adaptation of the group functionality. All of the already written docs could use this feature without any changes to the docs.

raboof: Completely agreed ;)

docs/src/main/paradox/features/snippet-inclusion.md
```diff
@@ -58,6 +58,45 @@ should not be highlighted set `type=text` in the directive's attribute section:
...
+
+The result will be rendered like this (try to switch tabs):
+
+First-java
+:   @@snip [example-first.java](../../resources/tab-switching/examples.java) { #java_first group=java }
+
+First-scala
+:   @@snip [example-first.scala](../../resources/tab-switching/examples.scala) { #scala_first group=scala }
+
+Some separator with text describing the @java[Java variant] @scala[Scala variant containing ***markdown*** and @ref:[Linking](linking.md)]
```
ktoso: great 👍

2m:
When merging, make sure to create a merge commit (as opposed to squash and merge), as this PR contains more than one author.

docs/src/main/paradox/features/snippet-inclusion.md
```diff
@@ -58,6 +58,45 @@ should not be highlighted set `type=text` in the directive's attribute section:
...
+
+Java
+:   @@snip [example-second.java](../../resources/tab-switching/examples.java)
+
+Scala
+:   @@snip [example-second.scala](../../resources/tab-switching/examples.scala)
```
2m: The example markdown does not match the markdown for render.

2m:
Its a pity I was not able to try it out, as paradox itself does not use the same code to generate the docs.

pvlugter:
I tried this out on the Cinnamon docs. The Cinnamon docs also uses tabs that shouldn't be switched. We have example `Configuration` snippets with a second tab called `Defaults` that displays the default configuration from a `reference.conf` ([for example](http://tech-hub.lightbend.com/docs/cinnamon/latest/plugins/chmetrics/reporters.html)). It would be good if the switching could be disabled for these. It would also be great to support tabs without snippets (with text or direct code examples). I also got javascript errors for a direct code example in a tab.

Given that we now have a declarative list of the switchable groups in the build, I'm wondering if we could use that as a whitelist for switching (by injecting it into the template for the javascript part). And instead of looking for classes on code blocks in tabs, could we simply use the tab name itself (and match against the known groups). So for the Cinnamon docs, we would define the switchable tabs as `"language" -> ("scala, "java), "build tool" -> ("sbt", "maven", "gradle")`. And the configuration tabs would be left as is. We could also store something like `language=java build-tool=gradle` in the cookie then, rather than appending the latest group.

raboof:
@pvlugter yes that makes a lot of sense to me!

eed3si9n:
What's the status on this PR? Is it ready to be merged?

raboof:
@eed3si9n I think it might make sense to implement the ideas by @pvlugter before merging, and perhaps add hooks to customize the behavior on tab change (as e.g. for the akka docs we might want to do a page refresh instead of client-side switching until we've migrated all pages to support that). @pvlugter WDYT?

eed3si9n:
ok. Sounds good.

pvlugter:
SGTM

pvlugter:
@raboof are you going to work on these changes?

raboof:
Yes if that doesn't interfere with your work I think I can look into it today!

pvlugter:
Great, sounds good.

raboof:
I've made some changes, including:
* only switch groups declared in the paradox config
* switch based on tab name when group property is missing (should be the common case)
* use 'neater' cookie format
* don't assume tabs contain snippets so heavily
* allow extending group switch javascript
* added documentation

I think this makes this PR ready for merge, review welcome.

Issue #101, opened 2017-03-20
Can not override page.st in src/main/paradox/_template


zoosky:
While css overriding works in `src/main/paradox/_template/css`, it seems not to work for `src/main/paradox/_template/page.st`

Paradox com.lightbend.paradox:paradox-theme-generic:0.2.9

JohnMurray:
I have also attempted to modify the `page.st` and `title.st` templates to no avail.

pvlugter:
I expect this was fixed in #108? There's also a test for overriding the page.st in the generic template. Available in paradox 0.2.11.

metasim:
I was able to do this with this tip provided by @jonas 

https://github.com/lightbend/paradox/issues/139#issuecomment-326136971

zoosky:
I tested with 0.3.0 and it works for me. :+1:  Thanks.

Issue #91, opened 2017-01-30
Feature: support `sbt ~paradox`


Sciss:
It would be great if paradox could track the source files (custom template, input markdown etc.) and automatically re-render when running as `sbt ~paradox` and one edits a file, e.g. `page.st`.

JohnMurray:
This is currently working for me for the input markdown. I can't speak to the custom template since I'm not using one ATM.

Sciss:
Yes, sorry, I think this was just for tuning the css files, the markdown files are traced indeed.

pvlugter:
Should be fixed in #108. Available in paradox 0.2.11.

PR #108, opened 2017-05-01
Remove global cache for templates to allow changes

pvlugter:
Having templates cached globally means that the template cannot be changed and the docs rebuilt in the same sbt process. So to change the template and rebuild, we need to keep restarting sbt / running in batch mode. Rather than invalidate the cache, I've just removed it completely. StringTemplate already provides caching for the template group, as far as I know. Also move the default template name to an sbt setting.

Also add an sbt-test that makes some changes to the template (and starts by overriding the generic template). This test will fail with the cached templates, without these changes.

Related issues: #91 and #101.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
+++ b/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
@@ -17,7 +17,7 @@
 package com.lightbend.paradox
 
 import com.lightbend.paradox.markdown.{ Breadcrumbs, Page, Path, Reader, TableOfContents, Writer, Frontin, PropertyUrl, Url }
-import com.lightbend.paradox.template.{ CachedTemplates, PageTemplate }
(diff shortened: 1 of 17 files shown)


pvlugter:
AppVeyor test is failing with some kind of encoding difference in a file that wasn't touched in this PR? 

`<title>A ? Paradox Test</title>` vs `<title>A � Paradox Test</title>`.

eed3si9n:
Sounds related to two of the PR that I just mentioned in https://github.com/lightbend/paradox/releases/tag/v0.2.10.

eed3si9n:
Opened https://github.com/lightbend/paradox/issues/111

eed3si9n:
Fixed it in https://github.com/lightbend/paradox/pull/112

pvlugter:
Cool. Thanks!

pvlugter:
I also bumped master to a snapshot version again for the sbt tests.

PR #114, opened 2017-05-10
Auto-expanding navigation

pvlugter:
Allow TOCs to expand for the active page. When enabled, the TOC will expand beyond the regular depth down to the active page, and also expand descendants of the active page down to the expand depth. Useful for a large or deep index, where having a full navigation doesn't work so well.

Add an option for including headers in the navigation. Also add `page` and `header` classes for the links in the TOCs, so they can be styled differently as needed.

This is being used in the conversion of Akka documentation to Paradox. WIP version of the docs that uses expanding navigation with headers can be found here:

https://downloads.lightbend.com/paradox/akka-docs-new/20170510-wip/scala.html


ktoso:
Whoa this looks awesome, thanks Peter :)

eed3si9n:
Looks awesome. This is going to be a game changer I think.

pvlugter:
Rebased on master for the merge conflicts. Merging now.

PR #117, opened 2017-05-17
fiddle fixes: string format, path config from snip

andreaTP:
Fixed a bug in string formatting and aligned code importing path to `snip` directive


andreaTP:
thanks! When you will release 0.2.12?

andreaTP:
ok, that was fast, thanks again!

PR #142, opened 2017-09-01
Add @@dependency directive

jonas:
Shows tabbed build tool snippets for using a JVM library.

---
Alpakka has quite a lot of these and Akka HTTP is starting to as well so it might be good to have a special directive for rendering this.

screen shot 2017-09-01 at 11 11 57 am

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -497,7 +497,8 @@ case class InlineGroupDirective(groups: Seq[String]) extends InlineDirective(gro
  * Dependency directive.
  */
 case class DependencyDirective(variables: Map[String, String]) extends LeafBlockDirective("dependency") {
-  val ScalaSuffix = List("_2.10", "_2.11", "_2.12")
+  val ScalaVersion = variables.get("scala.version")
+  val ScalaBinaryVersion = variables.get("scala.binary.version")
 
   def render(node: DirectiveNode, visitor: Visitor, printer: Printer): Unit = {
     node.contentsNode.getChildren.asScala.headOption match {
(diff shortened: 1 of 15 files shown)


jonas:
Akka HTTP also uses `cross CrossVersion.full`, so let me know if it would make sense to try to do this as part of the scala version suffix parsing.

jonas:
[Several places](https://github.com/akka/akka-http/search?utf8=✓&q=crossString&type=) when referring to the Akka HTTP artifacts. I guess it is for when it's compiled against a Scala RC version. Would it make sense to try to match Scala RC versions in the artifact name like [here](https://github.com/akka/akka-http/blob/master/build.sbt#L166-L169)?

2m:
I see. I was searching for a wrong thing.

I am wondering if we should match the Scala version from the build itself,
instead from the artifact name. It would save many search and replaces when
updating Scala version.

And yes, it does make sense to move the crossversion logic here.

2m:
I am curious about the default case:
```
@@dependency[sbt, Maven]{
  group=com.typesafe
  artifact=config
  version=1.3.1
}
```

For this example the sbt output would contain a single `%` thus with no cross-scala version support. Which is fine for this particular library, but it is not the most common case (scala libs).

What do you think if we would have `%%` in sbt and `_${scala.binary}` in others as a default? This would cover most of the Scala libraries. And for java libs we would need to have something special that would indicate that no cross scala version needs to be specified.

jonas:
I think it will be too error prone if we default to append the Scala binary version to the artifact name. In addition to having something special for java libs you'd also need something special when artifacts are versioned with `CrossVersion.full`.

For Scala dependencies this is pretty clear, yes it adds a bit of noise for the artifact, but it has less chance of doing the wrong thing:
```
@@dependency[sbt, Maven]{
  group=com.typesafe.akka
  artifact=akka-http_$scala.binary.version$
  version=$project.version$
}
```
The sbt snippet will show it as `"com.typesafe.akka" %% "akka-http" % "10.0.10"`

2m:
Great! I did not realise that we will be able to use `$...$` placeholders in the directive itself. Great stuff!

ktoso:
This looks great, LGTM; One problematic thing is that sometimes we need to show multiple dependencies and would be nice to do so in the same snippet. Perhaps that could be a new directive that would be "`@@dependencies`" I guess though

jonas:
Yes, in Akka HTTP there's one such case, I believe. @ktoso do you have any suggestions on how the syntax should be? I am not sure if the directive syntax parsing can be nested.

PR #143, opened 2017-09-10
Fix GitHub directive to support line numbers

jonas:
Reported in https://github.com/akka/akka-http/issues/1416 (and https://github.com/akka/akka-http/issues/900).


jonas:
I managed to pull off the solution mentioned here https://github.com/akka/akka-http/issues/1416#issuecomment-328279737 so it is possible to link to a snippet label and ensure the file exists and that line numbers don't get out of date.

```
@github[example](/src/test/scala/ServerSetup.scala) { #ssl-setup }
```

jonas:
One nice to have for the GitHub tree paths, is to check if the path exists iff in the GitHub tree URL (i.e. `/tree/$version`) matches `project.version` or points to `master`. Does this make sense?

PR #151, opened 2017-09-20
Bump paradox plugin to version 0.3.0

jonas:
Fixes the example display on the library dependency page.


jonas:
@bantonsson Självklart. Done.

PR #153, opened 2017-09-27
Disable autoScalaLibrary in theme projects

jonas:
Themes should not depend on Scala, since the artifacts are not cross
versioned.

---

Per discussion here: https://gitter.im/sbt/sbt-site?at=59cab44f614889d475415026

`com.lightbend.paradox:paradox-theme-generic:0.3.0` currently brings in Scala 2.12.

screen shot 2017-09-26 at 20 54 14


pvlugter:
Good catch. Surprised that it hasn't come up earlier. I'll publish a patch release for this now.

jonas:
Yes, now that the Paradox plugin is cross built for sbt 1.0 it may have an effect on projects using Scala 2.11. I am still curious why the theme ends up in the project classpath. Shouldn't it only live in the build project's classpath?

pvlugter:
The theme is packaged as a webjar, unpacked by sbt-web. But we could revisit this. Having it as a project level dependency seems generally useful though - like supporting multiple doc projects with different themes.

Published `0.3.1` release now.

PR #156, opened 2017-10-03
Allow site root relative base URLs in *doc & extref directives

bantonsson:
Not sure if this is the way to go, but I wanted to be able to link to APIs, and files that end up in the same site tree as the generated paradox site, so I added the possibility to have base URLs starting with `.../` that are converted to site root relative URLs, in the same way that the image links are converted.


bantonsson:
Anybody got any opinions on this PR?

PR #163, opened 2017-10-23
Clarify in readme who maintains

ktoso:


2m:
Umh. This duplicates the statement about Paradox support status. :) Let me fix that.

Issue #137, opened 2017-08-29
`@@@index` ToC does not include grouped page content


jonas:
In Akka HTTP, we are consolidating the Java and Scala documentation (akka/akka-http#1290) and I've noticed that the section overviews, which I believe is generated using the `@@@index` feature, does not include content in a page that have been grouped for example with

```
@@@ div { .group-java }
```

Some screenshots of JSON support:
<img width="777" alt="screen shot 2017-08-29 at 9 00 42 am" src="https://user-images.githubusercontent.com/8417/29822322-06370284-8c99-11e7-9c6b-eb2429264708.png">
<img width="807" alt="screen shot 2017-08-29 at 9 00 51 am" src="https://user-images.githubusercontent.com/8417/29822332-1244d362-8c99-11e7-8ea8-17d695dcefeb.png">

Before doc consolidation it was like [this](http://doc.akka.io/docs/akka-http/10.0.9/java/http/common/index.html):
<img width="776" alt="screen shot 2017-08-29 at 9 04 50 am" src="https://user-images.githubusercontent.com/8417/29822363-30837392-8c99-11e7-8f98-0d8d8162b592.png">

raboof:
Hmm, I wonder if this happens for any headings that are inside a `@@@ div` rather than at the 'top level'?

jonas:
Fixed by #170

Issue #128, opened 2017-06-19
Group support in ToC's


raboof:
Sometimes a whole section might exist in one group (e.g. 'scala' vs 'java') but not the other.

Would be nice if that could be encoded in the ToC so we can hide those as appropriate.

raboof:
On a related note: what if someone is at a page that, say, only exists in the Scala version. Should we still show the 'Java/Scala' dropdown? And if the user switches it, what should happen?

Hiding the Java/Scala dropdown might be the first intuitive choice, but when you think about it it's kind of problematic. Perhaps we should trigger a redirect to the parent page?

raboof:
I think this has been sufficiently covered by #170

PR #169, opened 2017-12-06
Allow 'group' property on pages

raboof:
Refs #128.

This is one experiment, looks promising, still untested.

Not sure this is the right way to go yet however, let me try another experiment that might also tackle #137 in one go.


raboof:
Closing in favour of #170

PR #170, opened 2017-12-06
Recognize headers within group blocks (#137)

raboof:
And add class tags to the index accordingly.

I think I prefer this approach to #169 

Should implement #137 and #128


raboof:
This appears to work well in #1600, @2m @pvlugter could you have a look?

jonas:
Does this also fix #128?

raboof:
@jonas yes it should! (tested with `routing-dsl/index.md` in https://github.com/akka/akka-http/pull/1600, this should have the 'Case Class Validation' section in Scala but not Java)

PR #172, opened 2017-12-12
Fix java URLEncoding akka #24099

andreaTP:
Fix for Akka https://github.com/akka/akka/issues/24099
Actually this was expected behavior, since scalafiddle is built on top of Akka-Http, probably something changed in the meanwhile in URL decoding in Akka HTTP itself (now is correct indeed).

cc. @ochrons @2m

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -381,6 +381,7 @@ case class FiddleDirective(page: Page, variables: Map[String, String])
           |  // $FiddleEnd
           |}
           |""".stripMargin, "UTF-8")
+        .replace("+", "%20") // due to: https://stackoverflow.com/questions/4737841/urlencoder-not-able-to-translate-space-character
 
       printer.println.print(s"""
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -381,6 +381,7 @@ case class FiddleDirective(page: Page, variables: Map[String, String])
           |  // $FiddleEnd
           |}
           |""".stripMargin, "UTF-8")
+        .replace("+", "%20") // due to: https://stackoverflow.com/questions/4737841/urlencoder-not-able-to-translate-space-character
```
raboof: I wonder if we could use some utilities from akka-http to build up the properly-formatted URL without having to use `URLEncoder` here, which is not really meant for this purpose

andreaTP: If you agree on having akka-http dependency to this sbt plugin it could be worth exploring this possibility of course

raboof: Ah sorry I thought for a moment we already had that dependency - let's not add it.

Not super nice but not sure introducing a dependency just for this is warranted, so 👍

PR #176, opened 2017-12-18
Speed up @ref directive by pre-calculating part of mappings

jrudolph:
This means that `relativeMapping` are calculated only once per run instead of for every invocation of a `@ref` directive.

For akka-http docs this gets ~30% total performance boost from 9-10 seconds to 6 seconds.


ktoso:
LGTM

PR #183, opened 2018-01-24
Use the current version in the addSbtPlugin example

raboof:
For a ready to copy-paste example


raboof:
Oh, just noticed that https://developer.lightbend.com/docs/paradox/latest/ appears to correspond to the lastest master, not to the last released version.

I guess we should introduce that (https://developer.lightbend.com/docs/paradox/current ?) before merging this PR.

2m:
Latest released version of paradox docs is published after https://github.com/lightbend/techhub/pull/196

Therefore we can merge this now.

Issue #185, opened 2018-01-25
Nested WrapDirective's


raboof:
It seems it is not possible to nest `WrapDirective`s. Is that a restriction of the pegdown parser or something we might be able to make possible?

jrudolph:
I think it might be that the parser is too strict when parsing container block directive content.

jrudolph:
I had a quick look into it but it doesn't seem easy to do. When the parser sees a starting `@@@ xys` it basically skims over the next text until it finds the next ending `@@@` and then recursively parses the contents afterwards. What we would have to do is to correspond matching pairs in the parser.

2m:
Fixed in https://github.com/lightbend/paradox/pull/188

PR #188, opened 2018-02-06
Add test and docs for nested block directives

raboof:
Refs #185


2m:
Does it work because of the additional `@`? If so, I think we should mention that explicitly in the docs.

raboof:
Yes that's what helps it match correctly

raboof:
done

2m:
Awesome pawsome! :feet:

pvlugter:
👍 thanks for documenting

PR #189, opened 2018-02-07
Allow adding classes to snippets

raboof:


core/src/test/scala/com/lightbend/paradox/markdown/SnipDirectiveSpec.scala
```diff
@@ -68,14 +68,18 @@ class SnipDirectiveSpec extends MarkdownBaseSpec {
...
     markdown("""
-      |@@@ div { #yeah .red .blue }
-      |Simple sentence here.
-      |@@@""") shouldEqual html("""
-      |<div id="yeah" class="red blue">
-      |<p>Simple sentence here.</p>
-      |</div>""")
+      |@@snip[example2.java](core/src/test/scala/com/lightbend/paradox/markdown/example2.java){ #example2 .red .blue }
+      |""") shouldEqual html("""
+      |<pre class="prettyprint red blue">
```
jonas: It would be nice to also propagate the `id` for completeness.

raboof: The confusing thing is the id is already also used to select the part of the file that needs to be included in the snippet. Would it make sense to also put it on the enclosing `pre` as id?

jonas: Oh, true. Didn't think about that. It could make sense, but on the other hand it is not necessarily unique in the package.

raboof: Right, probably safer to leave it be indeed

Issue #193, opened 2018-03-26
Task to open browser in docs -- ParadoxBrowse


ktoso:
We had this in Akka https://github.com/akka/akka/blob/master/project/ParadoxBrowse.scala

It would be nice to have it in core paradox as I expected this functionality across other projects and then realised its only in Akka :)

PR #194, opened 2018-03-26
add paradoxBrowse #193

ktoso:
refs #193

WDYT?


2m:
Does the css and javascript work as expected when opening the index html file like this? I always start a web server (`webfsd`) to serve the documentation from localhost when working on the docs.

ktoso:
Sure, nothing fancy in there after all; search of course won't work but that would not work on a locally started server anyway

PR #201, opened 2018-06-08
Fixup multisnippet indent

2m:
This adds intent tracking at the start of the block. This indent is then removed for every line of the block. The result is evenly indented multi-snippets even when the original code has different indentations for different snippets.


pvlugter:
Nice one

JustinPihony:
@2m is there a usecase for this which we can review when fixing up the overindentation issue?

2m:
The newly added test-case exercises that code: https://github.com/lightbend/paradox/pull/201/files#diff-8d79ec1d035f6c1d99a65339227c81a2R85

Also any multi-snippet that has different indentation, like [here](https://developer.lightbend.com/docs/alpakka/current/examples/jms-samples.html#example-read-text-messages-from-jms-queue-and-append-to-file). The imports and the code have different indentation, but the rendered result has same and correct indentation.

Issue #207, opened 2018-06-15
URLs with # are filtered from snippets


ennru:
In case snippets contain links with hashmarks (eg. Javadoc links) these are filtered and not show in the extracted snippet.

2m:
Duplicate of #149

Issue #149, opened 2017-09-18
Snippet inclusion removes some valid lines containing # characters


agemooij:
Hi

The `@@snip` feature appears to remove lines from snippets if they contain certain patterns related to `#` characters. It does this silently, without any warnings so users might not initially notice that their example snippets have been altered.

This is obviously related to support for nested snippets since it only apply to certain patterns of text that strongly resemble nested snippet markers.

Below is the test I've used so far to determine which lines will get filtered out of the snippet and which won't:

```
// Filtered out:
// 1 line with bare hash mark followed by a character: #a
// 2 line with bare hash mark followed by a character: abc#def
// 3 line with bare link with anchor: http://some.domain.com/foo#bar
// 4 line with markdown link with anchor: [IBM Cloud Object Storage](s3.md#bluemix-cloud-object-storage-with-s3-api )
//
// Not filtered out:
// 1 line with bare hash mark: #
// 2 line with quoted hash mark: "#"
// 3 line with quoted hash mark followed by a character: "#a"
// 4 line with quoted hash mark surrounded by characters: "abc#def"
// 5 line with bracket-quoted hash mark surrounded by characters: [abc#def]
// 6 line with bare link: http://some.domain.com/foo/bar
// 7 line with quoted link: http://some.domain.com/foo/bar
// 8 line with quoted link with anchor: "http://some.domain.com/foo#bar"
// 9 line with markdown link: [IBM Cloud Object Storage](s3.md)
// 10 line with quoted markdown link: "[IBM Cloud Object Storage](s3.md)"
// 11 line with quoted markdown link with anchor: "[IBM Cloud Object Storage](s3.md#bluemix-cloud-object-storage-with-s3-api)"
// 12 line with markdown link with anchor ending in bracket: [IBM Cloud Object Storage](s3.md#bluemix-cloud-object-storage-with-s3-api)
```

I'm not sure whether this can be entirely fixed but I think it is certainly counter-intuitive for most users that their links will be removed so perhaps the filter can at least try to detect links and URLs to prevent those from being filtered out?

/cc @olofwalker

pvlugter:
Yes, this is for nested snippets. Originally from https://github.com/lightbend/paradox/commit/a3000b874d04a0f7da0806dbc2ccb751be3a1f43. Main use case is having full examples that wrap nested snippets but exclude the snippet marker comments.

A couple of approaches so that we don't match the fragment part of links:

[...]

PR #208, opened 2018-06-15
Add pending test to illustrate URL filtering

ennru:
Illustrates #207


ennru:
With [the work-around](https://github.com/lightbend/paradox/issues/149#issuecomment-462615778) from #285 this seems redundant.

Issue #203, opened 2018-06-12
Warn when browsing outdated documentation


2m:
There should be a warning when browsing a version of the documentation that is not the latest released version.

Paradox could publish a "machine-readable" json file which would contain the version of the documentation. Then a javascript script could check this file under a "well-known" URL (using "current" instead of a version) and show a warning if the current version is not the latest one.

Related to https://github.com/lightbend/paradox/issues/105

PR #210, opened 2018-06-19
Add version warning

ennru:
This adds the generation of a `paradox.json` file containing the project's name and version, and reading that file from a specified URL to check against the version of the generated docs.
In case those versions differ a version warning is shown on every page.
The solution is quite crude but simple.


ennru:
Fixes #203

docs/src/main/paradox/features/version-warning.md
```diff
@@ -0,0 +1,12 @@
+Version warning
+---------------
+
+Paradox supports showing a warning when users browse documentation which is not the current released version. This is achieved by storing a JSON file (`paradox.json`) together with the generated site and specifying a stable URL to where the released version will be available.
+
+The built in theme (`generic`) contains Javascript to fetch the JSON file and compare the version with the version for which the documentation showing was generated. Whenever they differ, a warning text shows on every page offering a link to the released version's page.
+
+To use this functionality, add `project.url` to Paradox properties
+
+```$scala
```
2m: Is the `$` here accidental?

docs/src/main/paradox/features/version-warning.md
```diff
@@ -0,0 +1,12 @@
...
+---------------
+
+Paradox supports showing a warning when users browse documentation which is not the current released version. This is achieved by storing a JSON file (`paradox.json`) together with the generated site and specifying a stable URL to where the released version will be available.
+
+The built in theme (`generic`) contains Javascript to fetch the JSON file and compare the version with the version for which the documentation showing was generated. Whenever they differ, a warning text shows on every page offering a link to the released version's page.
+
+To use this functionality, add `project.url` to Paradox properties
+
+```scala
+paradoxProperties += ("project.url" -> "https://developer.lightbend.com/docs/paradox/current/")
```
2m: I see that the URL that is used for the `paradox.json` contains a double slash:

> The Same Origin Policy disallows reading the remote resource at https://doc.akka.io/docs/akka-stream-kafka/current//paradox.json

ennru: Yes, I wanted it to work in case the `project.url` doesn't end with a slash.

themes/generic/src/main/assets/js/warnOldVersion.js
```diff
@@ -0,0 +1,22 @@
+function initOldVersionWarnings($, thisVersion, projectUrl) {
+    var schemeLessUrl = projectUrl;
+    if (projectUrl.startsWith("http://")) projectUrl = schemeLessUrl.substring(5);
+    else if (projectUrl.startsWith("https://")) projectUrl = schemeLessUrl.substring(6);
+    const url = schemeLessUrl + "/paradox.json";
+    console.log("trying to read version data from " + url);
```
2m: I think this can go. When debuging it is possible to see if and where the request is sent in the `Network` tab.

themes/generic/src/main/assets/js/warnOldVersion.js
```diff
@@ -0,0 +1,22 @@
+function initOldVersionWarnings($, thisVersion, projectUrl) {
+    var schemeLessUrl = projectUrl;
```
2m: To avoid double slash at the end, you can make sure that `schemeLessUrl` ends with a slash by adding `.replace(/\/?$/, '/')`.

This is from https://stackoverflow.com/questions/11531363/javascript-jquery-add-trailing-slash-to-url-if-not-present

ennru:
current style

themes/generic/src/main/assets/page.st
```diff
@@ -145,5 +148,7 @@ $endif$
 <script type="text/javascript" src="$page.base$lib/prettify/prettify.js"></script>
 <script type="text/javascript" src="$page.base$lib/prettify/lang-scala.js"></script>
 <script type="text/javascript">jQuery(function(){window.prettyPrint && prettyPrint()});</script>
+<script type="text/javascript">jQuery(function(jq){initOldVersionWarnings(jq, '$page.properties.("project.version")$', '$page.properties.("project.url")$')});</script>
```
2m: When the `project.url` is not set in the properties, the `initOldVersionWarnings` function is still called with empty string as a parameter. And then `paradox.json` is looked up at the root of the host. It should not send the request if there is no project url set.

PR #212, opened 2018-06-21
Exception while trying to resolve link for github that contains placeholders

2m:
There is currently a regression in just released `0.3.4` where an exception is being thrown when a snippet path contains a placeholder. A workaround is to disable github links wtih `"snip.github_link" -> "false"` in `paradoxProperties`.


jonas:
Eeh, this also reveals a bug where we are not actually building anything on Travis. 😱 
Created #214

2m:
Pushed a fix.

PR #213, opened 2018-06-21
Make documentation on `ref` directive consistent.

metasim:
This PR changes one instance to `@ref:`, in sync with most (all?) of the other examples. Given the colon is optional, one approach should be chosen and consistent throughout to avoid user confusion.

IMO, I like the colon.

### DIFF
--- a/docs/src/main/paradox/features/markdown.md
+++ b/docs/src/main/paradox/features/markdown.md
@@ -8,4 +8,4 @@ We currently use the [Pegdown](https://github.com/sirthias/pegdown#introduction)
 Some aspects of our Markdown dialect worth noting are:
 
 * Bulleted lists must always be preceded by an empty line (otherwise they'll be considered to be a 'normal' part of the previous paragraph)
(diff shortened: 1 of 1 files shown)


2m:
Hm. It is not clear to me why `ref` directive needs a colon and other directives do not. I would prefer to have it consistent. Need to investigate a bit more why it is like so currently.

raboof:
I think all directives allow but not require a colon?

jonas:
I find that the `:` helps readability when viewing GitHub's rendering the markdown files.

metasim:
The colon is used in all the other examples for `@ref`, that's why I brought it up.

pvlugter:
The colon is optional. Most of the inline directives register with two names, with and without the colon.

https://github.com/lightbend/paradox/blob/master/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala#L119

As @jonas mentioned, the `:` version was added for readability on github — since the rest of the inline directive will render as a link.

metasim:
As a user, I think the documentation should use use one approach throughout. If there are two different ways, I would suspect people would think it implies different semantics, seeding confusion.

2m:
Alright. The argument for the better readability when rendered in Github sounds good. However for consistency of usage of `ref` directive it is still a long way to go:

```
~/p/paradox> git grep "ref\[" | wc -l
41
~/p/paradox> git grep "ref:\[" | wc -l
19
```

PRs are welcome!

PR #177, opened 2017-12-22
Bump sbt to version 1.0.4

jonas:

PR #214, opened 2018-06-21
Prefix the sbt command sequence with ';'

jonas:
Broken by me in #177 so that nothing was actually being built.
Noticeable in #212 which should fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
 language: scala
 jdk: oraclejdk8
 scala: 2.12.6
-script: sbt ";^^$TRAVIS_SBT_VERSION;verify"
-
-matrix:
-  include:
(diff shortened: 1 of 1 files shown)


jonas:
Looks like the combination of
```
addCommandAlias("verify", ";^ test:compile ;^ compile:doc ;^ test ;^ scripted ;docs/paradox")
```
and invoking it using
```
sbt ";^^$TRAVIS_SBT_VERSION;verify"
```
causes sbt to run the scripted tests for each of the configured versions in both builds in the build matrix.

2m:
Good point. I'll change the script to `sbt verify` so it is with parity with appveyor configuration.

jonas:
Thanks @2m, makes sense.

PR #226, opened 2018-07-20
Add the paradoxTheme to libraryDependencies using a special scope

### LINKED ISSUE #225: paradoxTheme added to libraryDependencies
With [solrs](http://inoio.github.io/solrs/) I just started to use paradox - thanks for this awesome tool!

Now I realized, that this added the `paradoxTheme` to the `libraryDependecies` (see [ParadoxPlugin.scala#L54](https://github.com/lightbend/paradox/blame/master/plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala#L54).
In consequence, this dependency is also written as dependency to the generated pom.xml on `publish`.

Appearently, a user of my library should not get the paradox theme as a transitive dependency.

This issue can be reproduced with `publishLocal` and inspecting the generated pom.xml.

Can this be fixed?

Note: this has already been asked in gitter [here](https://gitter.im/lightbend/paradox?at=5a90619de4ff28713aba3262), but was not answered AFAICS.

magro:
This shall prevent that `paradoxTheme` is added to the dependencies
of the project using paradox, which would mean that project users would
unexpectedly get the paradoxTheme as a transitive dependency.

Fixes #225


plugin/src/sbt-test/paradox/libraryDependency/test
```diff
@@ -0,0 +1,9 @@
+# check that the generated pom does not contain the paradoxTheme as dependency
+> publishLocal
+
+# ensure that our test is valid, check that this grep is successful
+$ exec grep "<artifactId>scala-library" target/scala-2.12/librarydependencytest_2.12-0.1.0-SNAPSHOT.pom
```
magro: Tests are failing because they're run with 2.10 AFAICS. Unfortunately I couldn't get it to work with wildcards - is there any way to write it in more generic way?

plugin/src/sbt-test/paradox/libraryDependency/build.sbt
```diff
@@ -0,0 +1,6 @@
+lazy val libraryDependencyTest = project
+  .in(file("."))
+  .enablePlugins(ParadoxPlugin)
+  .settings(
+    paradoxTheme := Some(builtinParadoxTheme("generic"))
```
eed3si9n: Could you use https://jonas.github.io/paradox-material-theme/ for testing please?

magro: Sure. Why?

eed3si9n: I just want to make sure that it's not picking up generic theme by accident.

magro: I have changed `check` to also catch the material theme and extended the test to test both the material theme and the generic theme. It fails for the material theme as well, with `assertion failed: pom contains paradox-theme dependency: Some(<artifactId>paradox-material-theme</artifactId>)`.

plugin/src/sbt-test/paradox/libraryDependencies/test
```diff
@@ -0,0 +1,14 @@
+> paradox
+
+# make sure that theme is retrieved
+$ exists target/paradox/site/main/assets/stylesheets/paradox-material-theme.css
+
+# check that the generated pom does not contain the paradoxTheme as dependency
+> makePom
```
2m: This is not needed as `packagedArtifact in makePom` is resolved from `check`.

plugin/src/sbt-test/paradox/libraryDependencies/build.sbt
```diff
@@ -0,0 +1,16 @@
+lazy val libraryDependencyTest = project
+  .in(file("."))
+  .enablePlugins(ParadoxPlugin)
+  .enablePlugins(ParadoxMaterialThemePlugin)
+  .settings(
+    // paradoxTheme := Some(builtinParadoxTheme("generic")),
```
2m: Lets remove this.

magro:
Thanks for merging! Is there a rough ETA for 0.3.6?

2m:
I have a couple of smallish fixes that I want to do for Alpakka. So I say lets aim for the release early next week?

magro:
Great, thanks in advance!

PR #227, opened 2018-07-24
Improve github.base_url derivation

2m:
This PR improves automatic `github.base_url` property derivation. If a project is on a non-snapshot version, it will set `github.base_url` to `.../tree/v<version>` instead of always to `.../tree/master`.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -259,7 +259,6 @@ case class JavadocDirective(page: Page, variables: Map[String, String])
 
 object GitHubResolver {
   val baseUrl = "github.base_url"
-
 }
 
 trait GitHubResolver {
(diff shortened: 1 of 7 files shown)


2m:
Fixed test when running on sbt 0.13.

jrudolph:
So, when releasing from a tagged version, the source links will point to the source file of the version (" The source code for this page can be found here. ")? What other feature makes use of the `github.base_url`?

It seems one drawback is that direct editing in Github doesn't work any more in that case because that's only allowed from branches. 

There's the general difficulty that a documentation page may already be outdated, either because you are looking at an old version or even the newest released version has been updated but not yet released. I wonder if we can actually do anything about that. For the source links, I somewhat think that the current behavior to link to the latest source version is actually fine because then the user is directly confronted with a potential difference between the published docs and its latest sources. WDYT?

2m:
> What other feature makes use of the `github.base_url`?

The github (`@github:[#1](#1)`) directive uses that as well to link to issues and commits. And it handles the `github.base_url` property with additional `/tree/...` part well.

> [...] then the user is directly confronted with a potential difference between the published docs and its latest sources.

It is fine if the changes are only in the file. But if the file has been moved, then the user gets a 404 from github (this was the original motivator for this change). Then the user needs to dig manually to discover where the file has been moved.

If, instead, a Github page of a tagged file is opened, then it is easier to navigate to the latest version.

jrudolph:
> If, instead, a Github page of a tagged file is opened, then it is easier to navigate to the latest version.

Ah, interesting so Github automatically follows renamed files when choosing another version from the dropdown?

jrudolph:
I see, sounds reasonable. LGTM

jrudolph:
(And thanks for the refined-github pointer, I wanted to try that for some time but never got to it)

PR #231, opened 2018-08-01
Improved scalafiddle integration

andreaTP:
Here we adopt an higher-level integration with scalafiddle leveraging: https://github.com/scalafiddle/scalafiddle-core/tree/master/integrations

This includes (mostly thanks to @ochrons ):
 - better and more readable tests
 - consistent offline code show (I recall we have an issue tracking this ...)
 - nicer UX
 - much less hacky interface

Note:
 - this completely break compatibility (since we are in `0.X` I guess it's fine)

cc. @ochrons @2m @patriknw @ktoso


andreaTP:
I will check consistency and update the Akka docs in the next days possibly (certainly before finalizing this PR)

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.3.6-SNAPSHOT"
+version in ThisBuild := "0.4.0-SNAPSHOT"
```
patriknw: what is the reason for the "big" bump?

andreaTP: the fiddle integration breaks the API

andreaTP:
I tested it publishing everything with akka-docs locally from the other PR branches, not sure how to publish everything ...

version.sbt
andreaTP: e.g. a `0.3` fiddle directive will no work with `0.4`

patriknw: ok, then the bump makes sense

ktoso:
LGTM, thanks

ktoso:
Give us time... soon.

Issue #191, opened 2018-03-20
Release from travis


raboof:
Right now releasing requires doing 'sbt release' with a PGP key that is correctly registered. Would be nice if all we had to do is create a tag through the github UI and let travis take care of the rest.

ennru:
Fixed with #256

PR #238, opened 2018-08-22
Document releasing

raboof:
Quick notes until we do https://github.com/lightbend/paradox/issues/191 :)


RELEASING.md
```diff
@@ -0,0 +1,3 @@
+* publishSigned for sbt 0.13 and 1.x
```
2m: Change to `^publishSigned` so it is clear how to publish for both sbt versions.

Issue #235, opened 2018-08-17
support JDK 10


xuwei-k:
pegdown depends on parboiled, parboiled depends on asm 5.2, asm 5.2 does not work with JDK10 😢 

https://github.com/sirthias/parboiled/blob/04277952b8df3f6b76067c5a456aece721ecc00c/project/Dependencies.scala#L15-L18

related to https://github.com/lightbend/paradox/issues/81

xuwei-k:
https://github.com/sirthias/parboiled/pull/123

SethTisue:
paradox is now green at https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk10-integrate-community-build/249/consoleFull :+1:

PR #240, opened 2018-08-27
jdk10 support

xuwei-k:
fix https://github.com/lightbend/paradox/issues/235

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
@@ -42,13 +42,9 @@ object Frontin {
 
   def loadProperties(str: Option[String]): Map[String, String] = str match {
     case None => Map.empty[String, String]
-    case _ =>
+    case Some(s) =>
       val p = new java.util.Properties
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
```diff
@@ -47,7 +47,7 @@ object Frontin {
       for (s <- str) {
         val q = new java.util.Properties
         q.load(new StringReader(s))
-        p.putAll(q)
+        (p: java.util.Hashtable[AnyRef, AnyRef]).putAll(q) // https://github.com/scala/bug/issues/10418
```
raboof: Since this function returns a `Map[String, String]`, perhaps we could replace this whole thing with
```
  case Some(s) => 
    val p = new java.util.Properties
    p.load(new StringReader(s)
    p.asScala.toMap
```
?

xuwei-k: done

SethTisue:
thank you Yoshida-san!

Issue #228, opened 2018-07-29
Code indentation is overindented


JustinPihony:
This seems to be a result of https://github.com/lightbend/paradox/pull/201 

Take a look at https://developer.lightbend.com/docs/reactive-platform/2.0/setup/setup-sbt.html and the multi-line code snippets - the lines past the first are indented almost out of view.

ennru:
The indentation is picked by looking at the #-mark, move the marker to the correct indentation and it will work out.

PR #242, opened 2018-09-04
Don't truncate actual characters out of snippets 

### LINKED ISSUE #241: Snippets truncated

johanandren:
Fixes #241

### DIFF
--- a/tests/src/test/scala/com/lightbend/paradox/markdown/SnipDirectiveSpec.scala
+++ b/tests/src/test/scala/com/lightbend/paradox/markdown/SnipDirectiveSpec.scala
@@ -93,6 +93,18 @@ class SnipDirectiveSpec extends MarkdownBaseSpec {
       |</pre>""")
   }
 
+  it should "not truncate snippets" in {
+    markdown("""@@snip[example.scala](tests/src/test/scala/com/lightbend/paradox/markdown/example.scala) { #multi-indented-example }""") shouldEqual html("""
+                                                                                                                                                      |<pre class="prettyprint">
+                                                                                                                                                      |<code class="language-scala">
+                                                                                                                                                      |object AnotherIndentedExample {
+                                                                                                                                                      |def rendered(): Unit = {
+                                                                                                                                                      |}
+                                                                                                                                                      |}
+                                                                                                                                                      |class AnotherClass</code>
+                                                                                                                                                      |</pre>""")
+  }
+
   it should "add link to source" in {
     implicit val context = writerContextWithProperties(
       "github.base_url" -> "https://github.com/lightbend/paradox/tree/v0.2.1",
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/com/lightbend/paradox/markdown/SnipDirectiveSpec.scala
```diff
@@ -93,6 +93,19 @@ class SnipDirectiveSpec extends MarkdownBaseSpec {
       |</pre>""")
   }
 
+  it should "not truncate snippets" in {
+    markdown("""@@snip[example.scala](tests/src/test/scala/com/lightbend/paradox/markdown/example.scala) { #multi-indented-example }""") shouldEqual html("""
```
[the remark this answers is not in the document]
johanandren: You mean the actual test? Whichever auto formatter this project uses.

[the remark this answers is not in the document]
jrudolph: I meant that the expectation is indented quite a lot.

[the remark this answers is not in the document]
johanandren: Actually, now I see the other ones aren't so it was probably IntelliJ crazy-indenting when I pasted it at some point. I'll fix.

ennru:
Relates to #228

tests/src/test/scala/com/lightbend/paradox/markdown/SnipDirectiveSpec.scala
```diff
@@ -93,6 +93,19 @@ class SnipDirectiveSpec extends MarkdownBaseSpec {
...
+      |<pre class="prettyprint">
+      |<code class="language-scala">
+      |object AnotherIndentedExample {
+      |def rendered(): Unit = {
```
jrudolph: @ennru Do you mean this, that the inner indentation is lost here?

johanandren: Yeah, I think best would be to base indentation off the indentation of the following non-empty line or something like that rather? But I don't think it is possible to get a solution that works for all use-cases without having some extra flags for deciding if to keep indent level or drop (the imports + main method contents vs "normal" code

johanandren: Not sure I want to jump into trying to fix that here though, I just want to make sure no actual characters except whitespace is truncated for now.

ennru: Yes, just a note on what this gives. Not cutting text is the best solution for now. @2m and I discussed some interplay of code formatters and this snippet cutting, but we didn't have this case.

PR #246, opened 2018-09-11
Add class on the link to the full source on GitHub

jonas:
Open for ideas of better naming.

The main motivation is that it would be nice to be able to style those links. Example from the Material theme:

screen shot 2018-09-11 at 12 12 28


ktoso:
This seems good 👍

PR #248, opened 2018-09-12
sbt 1.2.1, use sbt-dynver, Paradox 4.2

ennru:


version.sbt
```diff
@@ -1 +0,0 @@
-version in ThisBuild := "0.4.3-SNAPSHOT"
```
raboof: (checked that this file is indeed removed, not just cleared. Weird github UI.

Issue #247, opened 2018-09-12
Releasing fails because paradox-theme-generic is published twice


jrudolph:
`paradox-theme-generic` submodule doesn't use cross-building but when running `~publishSigned` it will be published twice which fails.

We should update release instructions what to do about that or (better) fix it in the build.

ennru:
Fixed with https://github.com/lightbend/paradox/pull/256

PR #256, opened 2018-09-22
Automate release from Travis CI

2m:
This adds Travis CI stage jobs that publish and release artifacts.


.travis.yml
```diff
@@ -1,8 +1,32 @@
...
+  include:
+    - stage: test
+      script: sbt verify
+      jdk: oraclejdk8
+
+    - script: sbt verify
+      jdk: oraclejdk10
+
+    - stage: publish
+      script: sbt ^publishSigned
```
jonas: Given #247 I think this needs to selectively `^core/publishSigned ^plugin/publishSigned ...`

2m: Ahh, thanks for reminding me of that. I have disabled `genericTheme` publishing for sbt `0.13`. So that should solve it.

2m:
Merging to test the release process out.

PR #260, opened 2018-10-06
Change links to snippet source from text to icon

2m:
From discussion in https://github.com/akka/akka-paradox/issues/36

2018-10-06-113956_924x296_scrot

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/AnchorLink.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/AnchorLink.scala
@@ -28,7 +28,7 @@ import scala.collection.JavaConverters._
 class AnchorLinkSerializer extends ToHtmlSerializerPlugin {
   def visit(node: Node, visitor: Visitor, printer: Printer): Boolean = node match {
     case anchor: AnchorLinkSuperNode =>
-      printer.print(s"""<a href="#${anchor.name}" name="${anchor.name}" class="anchor"><span class="anchor-link icon"></span></a>""")
+      printer.print(s"""<a href="#${anchor.name}" name="${anchor.name}" class="anchor"><span class="anchor-link"></span></a>""")
       anchor.getChildren.asScala.foreach(_.accept(visitor))
       true
     case _ => false
--- a/core/src/main/scala/com/lightbend/paradox/markdown/AnchorLink.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/AnchorLink.scala
@@ -28,7 +28,7 @@ import scala.collection.JavaConverters._
 class AnchorLinkSerializer extends ToHtmlSerializerPlugin {
   def visit(node: Node, visitor: Visitor, printer: Printer): Boolean = node match {
     case anchor: AnchorLinkSuperNode =>
-      printer.print(s"""<a href="#${anchor.name}" name="${anchor.name}" class="anchor"><span class="anchor-link"></span></a>""")
+      printer.print(s"""<a href="#${anchor.name}" name="${anchor.name}" class="anchor"><span class="anchor-link icon"></span></a>""")
       anchor.getChildren.asScala.foreach(_.accept(visitor))
       true
     case _ => false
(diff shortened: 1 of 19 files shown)


core/src/main/scala/com/lightbend/paradox/markdown/AnchorLink.scala
```diff
@@ -28,7 +28,7 @@ import scala.collection.JavaConverters._
 class AnchorLinkSerializer extends ToHtmlSerializerPlugin {
   def visit(node: Node, visitor: Visitor, printer: Printer): Boolean = node match {
     case anchor: AnchorLinkSuperNode =>
-      printer.print(s"""<a href="#${anchor.name}" name="${anchor.name}" class="anchor"><span class="anchor-link"></span></a>""")
+      printer.print(s"""<a href="#${anchor.name}" name="${anchor.name}" class="anchor"><span class="anchor-link icon"></span></a>""")
```
jonas: Is this necessary? It doesn't look related to snippets and the `ancher-link` class is already pretty specific.

2m: I extracted the `.icon` CSS class to be general so it could be reused here and on the snippet source links.

Should I have done it differently?

jonas: The properties can also be shared by listing the selectors:

```css
a.anchor .anchor-link, .icon {
    /* ... */
}
```

2m: Fair enough. Pushed a commit that reverts the changes to the scripted tests and now shares properties across selectors.

core/src/main/scala/com/lightbend/paradox/markdown/StyledVerbatim.scala
```diff
@@ -44,6 +44,14 @@ abstract class StyledVerbatimSerializer extends VerbatimSerializer {
     }
     printer.print(">")
 
+    node match {
+      case vgn: VerbatimGroupNode =>
+        vgn.getSourceUrl.ifPresent(sourceUrl => {
+          printer.print(s"""<a class="icon go-to-source" href="$sourceUrl" target="_blank" title="Go to snippet source"></a>""")
```
ennru: Ah, I like the "open in new tab".

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -361,12 +362,10 @@ case class SnipDirective(page: Page, variables: Map[String, String])
       val (text, snippetLang) = Snippet(file, labels)
       val lang = Option(node.attributes.value("type")).getOrElse(snippetLang)
       val group = Option(node.attributes.value("group")).getOrElse("")
-      new VerbatimGroupNode(text, lang, group, node.attributes.classes).accept(visitor)
-      if (variables.contains(GitHubResolver.baseUrl) &&
-        variables.getOrElse(SnipDirective.showGithubLinks, "false") == "true") {
-        val p = resolvePath(page, Path.toUnixStyleRootPath(file.getAbsolutePath), labels.headOption).base.normalize.toString
-        new ClassyLinkNode(p, "snippet-full-source github", new TextNode("Full source at GitHub")).accept(visitor)
```
ennru: Shouldn't the `snippet-full-source` class be removed again with this PR?

2m: I did not find a mention of `snippet-full-source` anywhere else in Paradox. So it seems it was not even used for styling or anything else.

jonas: I added it so that I could style the element in the paradox material theme. :+1: for adding something equivalent in the new version.

2m: Currently the new links have `class="icon go-to-source"`. Is that OK?

PR #272, opened 2018-11-06
Change to icon which renders on MacOS

2m:


ennru:
And it renders on Windows 10, as well.

PR #281, opened 2019-01-31
Implemented markdown includes

### LINKED ISSUE #195: Include (sections from) other .md files
It might be nice have an 'include' directive that pulls in content from another .md file - perhaps even one outside the paradox sources, to be able to pull in a fragment that is also in the README.md

jroper:
Fixes #195

Implemented markdown includes. It's very similar to snips, except that instead of outputting the content in a verbatim node that then gets rendered, it parses the content with the reader and then writes it
with a writer.

A new context is created for the writer to allow the included markdown file to include other markdown files and code snippets relative to itself. However, not a lot else about the context is modified, for example, custom properties for the included page aren't read. This might be desirable in some cases, and indesirable in others, not sure, I think it's best to wait and see what issues people report with it to see what other context should be modified.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -18,8 +18,9 @@ package com.lightbend.paradox.markdown
 
 import com.lightbend.paradox.tree.Tree.Location
 import java.io.{ File, FileNotFoundException }
-import java.util.Optional;
+import java.util.Optional
 
+import com.lightbend.paradox.tree.Tree
 import org.pegdown.ast._
 import org.pegdown.ast.DirectiveNode.Format._
 import org.pegdown.plugins.ToHtmlSerializerPlugin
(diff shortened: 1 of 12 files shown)


jroper:
Appveyor build is stuck loading the sbt project due to problems connecting to Maven/Ivy repositories (the sbt project is prompting for a user response).

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -694,3 +695,29 @@ case class DependencyDirective(variables: Map[String, String]) extends LeafBlock
...
+    val file = resolveFile("include", source, page, context.properties)
+    val (text, snippetLang) = Snippet(file, labels)
+    // I guess we could support multiple markup languages in future...
+    if (snippetLang != "md" && snippetLang != "markdown") {
+      throw IncludeDirective.IncludeFormatException(snippetLang)
+    }
+    val includeNode = context.reader.read(text)
+
+    // This location has no forest around it... which probably means that things like toc and navigation can't
+    // be rendered inside snippets, which I'm ok with.
```
pvlugter: Yep, agree. Can always come back to it later.

PR #285, opened 2019-02-12
Improved includes

jroper:
This improves the includes, so rather than the included markdown being parsed at the parent parents render time, it's parsed and then included in the AST straight after the parent page is parsed.

One consequence of this is that now the right hand page navigation works, since the headers can be extracted from the included markdown. I think it's also more robust all around.

In order to implement this, I had to create a new node, an `IncludeNode`, which doesn't get parsed (the include is still a `DirectiveNode`), but rather a post parse processing step converts the include `DirectiveNode`'s into `IncludeNode`'s, by parsing the included file. The `IncludeNode` then has the included files `RootNode` in it, along with the included path and file object. Then, a serializer plugin, when it encounters the `IncludeNode`, creates a new context based on the file that the node came from, and renders the nodes children into that - this ensures that included files can still have things like snips with paths relative to the included file.

Doing all this exposed an issue - since we expect included markdown files to include directives that may contain things that look like labels, it would have a problem with #149. So I made the snippet parser be able to turn on or off the filtering out of lines containing labels. The include directive turns this off by default, while the snip and other directives turn it on by default. Since that option now existed, I also added support to all of the above directives to configure it, via a `filterLabel` flag. This provides a work around, if not a fix, to #149.

### DIFF
--- a/tests/src/test/scala/com/lightbend/paradox/markdown/IncludeDirectiveSpec.scala
+++ b/tests/src/test/scala/com/lightbend/paradox/markdown/IncludeDirectiveSpec.scala
@@ -61,16 +61,46 @@ class IncludeDirectiveSpec extends MarkdownBaseSpec {
   it should "include headers from nested snippets in the toc" in {
     markdown(
       """
-        |@@toc { depth=2 }
+        |# Page heading
         |
-        |# Hello
+        |This text appears here to push down the toc so to ensure that the headers below
+        |calculation works.
+        |
+        |@@toc { depth=1 }
         |
         |@@include(tests/src/test/resources/headers.md)
+        |
+        |## Heading 3
+        |
         |""") should include(html("""
            |<div class="toc">
            |  <ul>
            |    <li><a href="test.html#heading-1" class="header">Heading 1</a></li>
            |    <li><a href="test.html#heading-2" class="header">Heading 2</a></li>
+           |    <li><a href="test.html#heading-3" class="header">Heading 3</a></li>
+           |  </ul>
+           |</div>"""))
+  }
+
+  it should "include headers from outer snippets in a nested toc" in {
+    markdown(
+      """
+        |# Page heading
+        |
+        |## Above toc
+        |
+        |@@include(tests/src/test/resources/toc.md)
+        |
+        |## Heading 1
+        |## Heading 2
+        |## Heading 3
+        |
+        |""") should include(html("""
+           |<div class="toc">
+           |  <ul>
+           |    <li><a href="test.html#heading-1" class="header">Heading 1</a></li>
+           |    <li><a href="test.html#heading-2" class="header">Heading 2</a></li>
+           |    <li><a href="test.html#heading-3" class="header">Heading 3</a></li>
            |  </ul>
            |</div>"""))
   }
--- a/tests/src/test/scala/com/lightbend/paradox/markdown/IncludeDirectiveSpec.scala
+++ b/tests/src/test/scala/com/lightbend/paradox/markdown/IncludeDirectiveSpec.scala
@@ -50,4 +50,29 @@ class IncludeDirectiveSpec extends MarkdownBaseSpec {
       |<p>This file should be included by IncludeDirectiveSpec</p>""")
   }
 
+  it should "include nested code snippets" in {
+    markdown("""@@include(tests/src/test/resources/include-code-snip.md)""") shouldEqual html("""
+      |<pre class="prettyprint">
+      |<code class="language-conf">
+      |a = b</code>
+      |</pre>""")
+  }
+
+  it should "include headers from nested snippets in the toc" in {
+    markdown(
+      """
+        |@@toc { depth=2 }
+        |
+        |# Hello
+        |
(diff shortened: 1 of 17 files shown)


tests/src/test/scala/com/lightbend/paradox/markdown/IncludeDirectiveSpec.scala
```diff
@@ -50,4 +50,29 @@ class IncludeDirectiveSpec extends MarkdownBaseSpec {
...
+        |# Hello
+        |
+        |@@include(tests/src/test/resources/headers.md)
+        |""") should include(html("""
+           |<div class="toc">
+           |  <ul>
+           |    <li><a href="test.html#heading-1" class="header">Heading 1</a></li>
+           |    <li><a href="test.html#heading-2" class="header">Heading 2</a></li>
+           |  </ul>
+           |</div>"""))
```
2m: Is it OK, that `# Hello` is not included in the toc?

jroper: No, it's not, because Paradox strips the first header that it finds in the markdown from the page (regardless of where it is in the page) and treats that as the page title and heading. The behaviour is a bit weird when you don't put it right at the top, but all pages should start with a header so it's not a problem in practice.

That said, your comment caused me to look deeper into the handling, and I realised that toc directives (not navigation indexes) are quite broken because when they calculate if a header should be included in the page or not, they look at whether its index in the file is below the index of the toc directive. Of course, if the header is in an included file, or the toc is in an included file and the header is in the outer file, what needs to be compared is the include index to the header or toc index, since their locations in different files are not comparable. So, I've pushed a commit that fixes this - it includes a path of include indexes with every header, and adds the include index path to the rendering context, and the calculation has been updated to use this path in its considerations.

jroper:
@pvlugter I've now fixed the serialization of toc directives to work as someone might intuitively expect - it's able to do calculations of whether the header is below the toc directive correctly in the resulting page (by tracking a path of included indexes, where each index in the path is the index of the include directive in the parent file, both in the writer context and in the header refs, so essentially the location of the toc and the header ends up as two vectors of indexes that can be lexicographically compared to work out if one is below the other), and also I fully populated the tree when writing include files so toc directives in included files can work.

pvlugter:
@jroper Great that it fully supports toc directives now. I think the ideal could be that toc doesn't know about includes—by having fully expanded includes in the node tree or something—but the include index path approach looks good to me and makes sense 👍

pvlugter:
On the filter labels (and resolving #149) what about we allow the default to be configurable? Have a paradox property that allows setting whether filter labels is on or off by default, so people can flip it around if they want (don't filter by default, explicitly enable filtering). Can be set in sbt build or page properties. Going another step could be adding a configurable filter string or regex — so that people can disable filter labels by default and then add a `filterLabelsWith="!exclude-from-snippet!"` to have special filter text. Even though the default behaviour would be the same, seems like enough flexibility to consider it fixed?

jroper:
The includes are fully expanded in the node tree, the challenge is that the toc directive uses the location in the file to determine whether an element is above or below it, but if that's the mechanism, then location in file is not enough when you support includes, you need to know the location in which file, and where that file appears in the overall document. So the problem is that the toc directive uses a mechanism that inherently depends on files and locations in files, which means it must understand includes. There are two ways to resolve that, one is to offset all the locations in the indexes in the tree according to includes (this would have to not just apply to included files, but to the parent files as well since their locations need to be bumped down after the includes), but that would likely have other undesirable effects since now the indexes don't correlate to actual file locations, which if relied on for error reporting, would be a problem. The other way to resolve it would be to make the toc directive use the tree itself to determine whether an element was before or after it, this would mean not using the extracted headers, and instead walking the tree, finding itself, and then taking every header after that. That's probably the more robust solution, but it completely changes the way it works, I felt this was a simpler change. Plus, having the include index path in the writer context may potentially be useful in future, particularly for error reporting.

pvlugter:
Right, I was thinking about index rewriting as part of the expansion. But yeah, this is already a different approach to walking the trees — I think just because it was easier to quickly compare indexes, and given that it's taking the toc in the node tree to find a location in the separate header structure. So agree that this is the best way right now. Maybe we revisit approaches if/when there's a switch over to flexmark.

jroper:
Ok, I've added a `include.filterLabels`, `snip.filterLabels` and `fiddle.filterLabels` keys that can be configured globally or in the front matter of a file. The default behaviour doesn't change.

jroper:
I think any further mechanisms for configuring filtering labels should be a separate change, there's probably a lot to consider in how it should be implemented given that the current mechanism isn't regex based, but rather is on the snip directives existing heuristics for whether something is a label or not.

jroper:
Builds failed due to `Option.contains` not existing in Scala 2.10, replaced with `.exists(_ == ...)`.

pvlugter:
Sounds good. LGTM.

jroper:
Any objections to this being merged?

pvlugter:
👍

PR #289, opened 2019-03-07
Introduce canonical.base_url and docs for the description tag

ennru:


build.sbt
```diff
@@ -129,7 +129,9 @@ lazy val docs = (project in file("docs"))
     paradoxTheme := Some(builtinParadoxTheme("generic")),
     paradoxProperties in Compile ++= Map(
       "empty" -> "",
-      "version" -> version.value
+      "version" -> version.value,
+      "project.url" -> "https://developer.lightbend.com/docs/paradox/current/",
+      "canonical.base_url" -> "https://developer.lightbend.com/docs/paradox/current/",
```
2m: Since in most cases these are going to be the same, I say lets have it default to `project.url` here: https://github.com/lightbend/paradox/blob/master/plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala#L89

Also, a good default value for `project.url` is the value of `homepage` sbt setting key.

That would keep the duplication to the minimum, and will do the right thing out of the box.

ennru:
Right, amended this to default `canonical.base_url` will do the defaulting of `project.url` in another PR.

plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala
```diff
@@ -90,9 +90,14 @@ object ParadoxPlugin extends AutoPlugin {
       "snip.root.base_dir" -> baseDirectory.value.toString,
       SnipDirective.buildBaseDir -> (baseDirectory in ThisBuild).value.toString,
       SnipDirective.showGithubLinks -> "true",
+      "canonical.base_dir" -> (homepage in ThisBuild).value.toString,
```
2m: This one should go.

PR #304, opened 2019-03-22
Drop sbt 0.13 support

### LINKED ISSUE #291: Drop sbt 0.13 cross-build
Do we still require sbt 0.13 support for Paradox?

dwijnand:
... as well as:

 * drop Scala 2.10 support,
 * upgrade to Scala 2.12.8, and
 * upgrade to sbt 1.2.8

Fixes #291

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -32,23 +32,9 @@ object Common extends AutoPlugin {
 
   // AutomateHeaderPlugin is not an allRequirements-AutoPlugin, so explicitly add settings here:
   override def projectSettings = AutomateHeaderPlugin.projectSettings ++ Seq(
-    scalaVersion := { (sbtBinaryVersion in pluginCrossBuild).value match {
-      case "0.13" => "2.10.7"
-      case _ => "2.12.6"
-    }},
-    crossSbtVersions := Seq("0.13.17", "1.0.4"),
-    // fixed in https://github.com/sbt/sbt/pull/3397 (for sbt 0.13.17)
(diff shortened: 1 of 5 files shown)


project/Common.scala
```diff
@@ -32,23 +32,10 @@ object Common extends AutoPlugin {
...
   override def projectSettings = AutomateHeaderPlugin.projectSettings ++ Seq(
-    scalaVersion := { (sbtBinaryVersion in pluginCrossBuild).value match {
-      case "0.13" => "2.10.7"
-      case _ => "2.12.6"
-    }},
-    crossSbtVersions := Seq("0.13.17", "1.0.4"),
-    // fixed in https://github.com/sbt/sbt/pull/3397 (for sbt 0.13.17)
-    sbtBinaryVersion in update := (sbtBinaryVersion in pluginCrossBuild).value,
+    scalaVersion := "2.12.8",
+    crossSbtVersions := Seq("1.0.4"),
```
2m: Is this sbt version going to be used when running tests if we do not use `^` anymore?

dwijnand: No, `sbtVersion in pluginCross` (IIRC) will be used (which default to -> `sbtVersion` -> `sbt.version`).

Shall I drop the `crossSbtVersions` or keep the `^`?

2m: Setting `crossSbtVersions` to the last bugfix release in the series and using `^` makes sure that released plugin artifacts are compatible with older sbt versions. So the decision is about whether we want to still support older sbt 1.0.x and 1.1.x or not.

I think we can drop it, because if you are already using 1.x then it is quite easy to upgrade to 1.2.8.

Issue #316, opened 2019-05-24
canonical links on pages in subdirectories that are not part of any index don't work


raboof:
When creating canonical links is enabled, it seems spurious `../` are added to the URL when the page is not in the root of the documentation, and also not in any index.

Fairly isolated example at https://github.com/lightbend/paradox/pull/315/files#diff-4de2d28c71ba5ec414318b7071ce2586R9

PR #317, opened 2019-05-24
Throw an error when there are more ToC roots than expected

raboof:
As an unexpected extra ToC root likely indicates a mistake.

Should make it easier to find inconsistencies like those
that triggered https://github.com/lightbend/paradox/issues/316

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
+++ b/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
@@ -28,8 +28,6 @@ import org.stringtemplate.v4.STErrorListener
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
```diff
@@ -28,6 +28,8 @@ import org.stringtemplate.v4.STErrorListener
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
+import org.stringtemplate.v4.misc.STMessage
+import org.stringtemplate.v4.misc.ErrorType
```
2m: Not used?

raboof: good catch, removed

PR #326, opened 2019-06-26
Check internal references

### LINKED ISSUE #120: Check anchor targets in @ref directive
`@ref` already checks for valid page references. It also allows targets to be specified with `@ref:[page](page.md#target)`. It would be useful if it also checked that the target is a valid anchor in that page. Particularly as a heading could be updated without realising there are links to it.

A bigger question: should we even have explicit `@ref`s? Paradox could also check the validity of all relative links, and transform source md links to output html links automatically. And optionally check external links exist.

raboof:
Fixes #120


raboof:
Tested with Akka, found some invalid links, no false positives.

Issue #53, opened 2016-11-07
Doc validation


jroper:
The paradox sbt plugin should provide a task for validating docs.  This should include:

* Missing links tests
* Missing fragment links tests
* Missing scaladoc/javadoc links tests
* Missing scaladoc/javadoc fragment (ie, to specific methods/fields) links tests
* Missing code snippets tests
* Missing resources (eg, inline image links, etc) tests

An additional task should be provided that checks for dead external links - this should be an additional task, since a task that calls out to many external websites will cause problems if run from CI, both in terms of false negatives and also it's not playing nicely on the web.  Instead the intent would be to run this task manually every so often.

jonas:
For linking between Paradox sites, I suggest we consider packaging docs in JARs (which might already be the case for tech hub) so that this can be managed partially via dependencies. Checking links would be a matter of reading a manifest file which lists all pages and page anchors.

This would also allow downstream projects to provide Paradox properties such as where the Paradox site is hosted and Scaladoc/Javadoc base URLs.

benmccann:
This functionality can be found in [LightbendMarkdownValidation](https://github.com/typesafehub/lightbend-markdown/blob/master/plugin/src/main/scala/com/lightbend/markdown/sbt/LightbendMarkdownValidation.scala) and [PlayDocsValidation](https://github.com/playframework/playframework/blob/master/framework/src/play-docs-sbt-plugin/src/main/scala/com/typesafe/play/docs/sbtplugin/PlayDocsValidation.scala), so we can use those as a starting point. This is a pre-requisite for Play and/or Lagom moving to Paradox one day. The validation on Play and Lagom's docs catches a lot of errors during development.

2m:
Links to scaladoc with fragment syntax can be checked by sbt-sdlc: https://github.com/sbt/sbt-sdlc

jroper:
Validation for links to any assets that are available to the build (could include images, api docs, etc) has been implemented in #359, additionally that PR also allows optionally validating external links by making HTTP requests on them. I think this validation, in combination with the validation that paradox already offered as part of the compile task, gives all the functionality that I was hoping for (and more).

[...]

PR #328, opened 2019-06-27
Check for duplicate anchors

raboof:
Part of #53
Tested with https://github.com/akka/akka/pull/27226, needed to remove some 'manual' anchors but no further false positives


raboof:
superseded by #329

PR #329, opened 2019-06-27
Check for duplicate anchors

### LINKED ISSUE #319: Add check for dups in TOC
If there are any header duplicates on the page it will generate the TOC with the duplicates. It will also highlight all the duplicated items during navigation. We can improve it by making Paradox complain about any duplicated headers on the page.

raboof:
Part of #53. Resolves #319.
Tested with akka/akka#27226, needed to remove some 'manual' anchors but no further false positives


jrudolph:
Can we make these warnings and also not stop processing after the first occurrence?

PR #341, opened 2019-08-07
Warn rather than throw on duplicate anchors

raboof:
Refs #329


pvlugter:
👍 And for having "real reporting" at some point too.

PR #343, opened 2019-08-08
Allow checking the value of a property in a template

raboof:


raboof:
> Add to the documentation?
done

PR #345, opened 2019-08-12
FilterLabels: off for unlabeled includes

ennru:
## Purpose

Disable label filtering for snippets/includes/fiddles which include a complete file.
Document the `filterLabels` flag and property. 

## Background 

Especially for snippets including whole HOCON files, label filtering often hides important lines with important information as `#` is used for comments.

One example is the `reference.conf` snippet in https://doc.akka.io/docs/alpakka/1.1.1/dynamodb.html#setup

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -393,7 +393,7 @@ case class SnipDirective(page: Page, variables: Map[String, String])
     try {
       val labels = node.attributes.values("identifier").asScala
       val source = resolvedSource(node, page)
-      val filterLabels = node.attributes.booleanValue("filterLabels", variables.get("snip.filterLabels").forall(_ == "true"))
+      val filterLabels = labels.nonEmpty && node.attributes.booleanValue("filterLabels", variables.get("snip.filterLabels").forall(_ == "true"))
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -393,7 +393,7 @@ case class SnipDirective(page: Page, variables: Map[String, String])
     try {
       val labels = node.attributes.values("identifier").asScala
       val source = resolvedSource(node, page)
-      val filterLabels = node.attributes.booleanValue("filterLabels", variables.get("snip.filterLabels").forall(_ == "true"))
+      val filterLabels = labels.nonEmpty && node.attributes.booleanValue("filterLabels", variables.get("snip.filterLabels").forall(_ == "true"))
```
raboof: Could we allow filtering out the labels when including a full file? Like:
* if a `filterLabels` attribute is present, use it
* if not, and a `snip.filterLabels` default it set, use it
* else, `labels.nonEmpty`
?

(do you have a use for `snip.filterLabels`? might be simpler without that option)

ennru: I made the local attribute take precedence and if nothing is defined `labels.nonEmpty` and the attribute control it.
(The global setting was there before, I didn't want to drop it.)

PR #355, opened 2019-09-04
Made javadoc link style configurable

### LINKED ISSUE #354: Javadoc directive incompatible with JDK 9+
The javadoc directive links to the frames version of the docs, using the query parameter to pass the path to the class name, which is fine for javadocs generated with JDK8 and below. But JDK9's javadoc tool no longer has frames.

I propose we make it configurable as to which style of linking is used - possibly per package - and default to linking directly to the html page if paradox is being built with JDK9+. Although the javadocs aren't necessarily built by the same version of Java as paradox, in many cases, it will be part of the same build with the same version of Java, so it's a good default.

jroper:
Fixes #354

This allows generating javadoc links to javadocs generated by JDK9+.

### DIFF
--- a/docs/src/main/paradox/directives/linking.md
+++ b/docs/src/main/paradox/directives/linking.md
@@ -97,7 +97,15 @@ For example, given:
  - `javadoc.akka.http.base_url=http://doc.akka.io/japi/akka-http/10.0.0`
 
 Then `@javadoc[Http](akka.http.javadsl.Http#shutdownAllConnectionPools--)` will resolve to
-<http://doc.akka.io/japi/akka-http/10.0.0/akka/http/javadsl/Http.html#shutdownAllConnectionPools-->.
+<http://doc.akka.io/japi/akka-http/10.0.0/?akka/http/javadsl/Http.html#shutdownAllConnectionPools-->.
+
+The `@javadoc` directive offers two styles of linking, linking to the frames version of
+the apidocs, and linking to the non frames version. This can be controlled using the `javadoc.link_style`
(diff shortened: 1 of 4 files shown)


docs/src/main/paradox/directives/linking.md
```diff
@@ -97,7 +97,14 @@ For example, given:
  - `javadoc.akka.http.base_url=http://doc.akka.io/japi/akka-http/10.0.0`
 
 Then `@javadoc[Http](akka.http.javadsl.Http#shutdownAllConnectionPools--)` will resolve to
-<http://doc.akka.io/japi/akka-http/10.0.0/akka/http/javadsl/Http.html#shutdownAllConnectionPools-->.
+<http://doc.akka.io/japi/akka-http/10.0.0/?akka/http/javadsl/Http.html#shutdownAllConnectionPools-->.
+
+The `@javadoc` directive offers two styles of linking, linking to the frames version of
+the apidocs, and linking to the non frames version. This can be controlled using the `javadoc.link_style`
+property, setting it to `frames` for frames style linking, and `no-frames` for no frames style linking.
```
ennru: I wonder if `no-frames` is the best we can come up with. Maybe `page` would work? But this implementation accepts everything that is not `frames`, right?

jroper: Correct. I'll change to page.

jroper: I updated to `direct`, because it links direct to the page, instead of through the frames.

PR #363, opened 2019-09-10
Ensure URL fragments to named anchors also validate

jroper:


pvlugter:
The failure in the CI build is https://github.com/sbt/sbt/issues/5049. Not sure why it's using different versions in scripted, but I assume that means that if sbt-paradox is published with sbt 1.3.0 it no longer works in sbt 1.2.8 projects?

raboof:
Hmm, though that would also prevent projects like Akka gRPC, that themselves produce sbt plugins and might not be ready to require sbt 1.3.0, from upgrading. WDYT?

jroper:
Can't we just change the version of sbt that we build against? There's no reason that sbt plugins need to be built against the same version of sbt that they use to build themselves. So we should just set `sbtVersion` to the lowest version of sbt that we're compatible with, eg `1.0.0` if we're compatible with that.

raboof:
> There's no reason that sbt plugins need to be built against the same version of sbt that they use to build themselves

You're right of course. Then I'd be fine with just sticking with sbt 1.3.0 for paradox and setting `sbtVersion` in any downstream projects that need to.

pvlugter:
Another fragment check that doesn't work is to the datadog docs:

https://docs.datadoghq.com/tracing/guide/trace_sampling_and_storage/?tab=java#trace-storage

Looks like the id attribute is not quoted in this case: `<h2 id=trace-storage>Trace Storage</h2>`

pvlugter:
And single quotes of id/name attributes would also not work, assuming that's used and valid.

jroper:
Hmm... so we can keep coming up with edge cases etc and adding support for them, or we could do it properly, by parsing the document and then looking for named anchors or ids. This would mean using HtmlUnit (I'm not aware of any other HTML parser available on the JVM - there are plenty of xml parsers of course, but many web pages, such as the datadog docs example, are not valid XML so can't be parsed by an XML parser). Thoughts?

jroper:
I forgot about [jsoup](https://jsoup.org/).

pvlugter:
Yes, my next thought as well. Switching over to jsoup sounds good to me.

jroper:
Ok, I added jsoup. Also, I made a small improvement, links to the same page with different fragments only result in that page being downloaded once. Plus the code around grouping common links is simpler now.

pvlugter:
Cool, I'll try it out again.

core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
```diff
@@ -189,9 +205,13 @@ class ParadoxProcessor(reader: Reader = new Reader, writer: Writer = new Writer)
     }
   }
 
-  private def validateFragment(path: String, content: String, fragment: String, reportError: String => Unit) = {
-    if (!content.contains("id=\"" + fragment + "\"")) {
-      reportError(s"Could not find anchor id [$fragment] in page [$path]")
+  private def validateFragments(path: String, content: Document, fragments: List[CapturedLinkFragment], errorContext: ErrorContext): Unit = {
+    fragments.foreach {
+      case CapturedLinkFragment(Some(fragment), sources) =>
+        if (content.getElementById(fragment) == null && content.select(s"a[name=$fragment]").isEmpty) {
```
jroper: Wasn't sure what the best way to lookup named a tags was. Could probably lookup all a tags, and then iterate through to see if any have a name of fragment. This is simpler, though with the possibility of escaping issues, for example, it definitely won't work for fragments with `]` in them, there may also be other characters that will cause a problem. People don't usually put special characters that won't work in names since they're meant to appear in URLs, and URLs with percent encoding looks ugly, so it may just not be an issue. I think if we find any actual problems with this, we can switch to something that searches for it better.

jroper:
Yeah you're right, it is. To validate that we'd have to use htmlunit (which is likely to cause bigger problems because its javascript support is only partial) or selenium/webdriver (slow). I don't think it's worth it, I'd say just add an ignore (paradoxIgnorePaths).

PR #364, opened 2019-09-10
Update sbt in generated-source scripted test

raboof:


plugin/src/sbt-test/paradox/generated-source/project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.2.8
+sbt.version=1.3.0
```
jroper: Wouldn't it be better to just delete this file? I don't think it's needed, or wanted.

raboof: I have no preference

jroper: I think it should be deleted, otherwise we'll have to manually update it every time we upgrade sbt. No other scripted tests have it.

raboof: actually there are 2 more - updated

jroper:
@pvlugter agreed on that, I think that could be achieved by doing something like this:

```
sbt 'set scriptedSbt = "1.2.8"' scripted
```

PR #368, opened 2019-09-12
Fixed issues relating to JDK10 anchor link encoding

jroper:
In JDK10 onwards, the way that anchor links for methods are encoded has changed, compare:

```
https://docs.oracle.com/javase/10/docs/api/java/util/Map.html#put(K,V)
 https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#put-K-V-
```

This adds functionality to paradox to automatically convert between the two styles, by saying `javadoc.javadoc_version = 11` for example. The setting applies per path prefix, and if unset, no conversion is done. A special value of `detect` is supported, to allow converting based on the builds java version.

### DIFF
--- a/core/src/main/java/org/pegdown/ParserWithDirectives.java
+++ b/core/src/main/java/org/pegdown/ParserWithDirectives.java
@@ -18,6 +18,7 @@ package org.pegdown;
 
 import org.parboiled.Context;
 import org.parboiled.Rule;
+import org.parboiled.annotations.Cached;
 import org.parboiled.support.StringBuilderVar;
 import org.parboiled.support.Var;
 import org.pegdown.ast.*;
@@ -251,7 +252,11 @@ public class ParserWithDirectives extends Parser {
     }
 
     public Rule DirectDirectiveSource() {
-        return Sequence(Sp(), Enclosed('(', ANY, ')'), push(new DirectiveNode.Source.Direct(popAsString())));
+        return Sequence(
+                Sp(),
+                Enclosed('(', DirectiveLitChar(), ')'),
+                push(new DirectiveNode.Source.Direct(popAsString().replaceAll("\\\\(.)", "$1")))
+        );
     }
 
     public Rule RefDirectiveSource() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/pegdown/ParserWithDirectives.java
```diff
@@ -251,7 +252,11 @@ public Rule DirectiveSource() {
     }
 
     public Rule DirectDirectiveSource() {
-        return Sequence(Sp(), Enclosed('(', ANY, ')'), push(new DirectiveNode.Source.Direct(popAsString())));
+        return Sequence(
+                Sp(),
+                Enclosed('(', DirectiveLitChar(), ')'),
+                push(new DirectiveNode.Source.Direct(popAsString().replaceAll("\\\\(.)", "$1")))
+        );
```
jroper: Not sure if this is the best way to fix this, this is to allow you to put paranthesis in URLs, eg:

```
@javadoc[example](com.Example#example\(\))
```

I wasn't really in the mood to try and truly understand parboiled, this worked so I used it.

jroper:
Note also, one reason why conversion is desirable is because links to JDK10+ methods suck without it, you have to slash escape the parenthesis, eg:

```
@javadoc[example](com.Example#example\(\))
```

No one wants to write that, this is much nicer:

```
@javadoc[example](com.Example#example--)
```

docs/src/main/paradox/directives/linking.md
```diff
@@ -104,8 +104,20 @@ the apidocs, and linking to the non frames version. This can be controlled using
 property, setting it to `frames` for frames style linking, where rendered links link to the frames
 version of the javadocs, passing the class in the query parameter, and `direct` for linking direct
 to the classes page. The link style defaults to `direct` if the `java.specification.version` system
-property indicates Java 9+, since the JDK 9 Javadoc tool does not generate framed api docs. Otherwise,
-for Java 1.8 and earlier, it defaults to `frames`.
+property indicates Java 11+, since by in JDK 11, the Javadoc tool does not generate framed api docs. 
+Otherwise, for Java 10 and earlier, it defaults to `frames`.
```
raboof: Linking to frames from 11+ instead of 9+ seems like an independent change - OK if you're confident it's the right thing though :D

docs/src/main/paradox/directives/linking.md
```diff
@@ -104,8 +104,20 @@ the apidocs, and linking to the non frames version. This can be controlled using
...
+
+From JDK 10 onwards, anchors for Javadoc method references do not have `(`, `)` and `,` characters
+replaced with `-`. To facilitate in migrating, Paradox supports automatically converting between
```
raboof: Isn't it the other way around?

ennru:
With the original problem of parens in the directive solved by #386 I suggest we close this.
@jroper objections?

PR #370, opened 2019-09-13
PDF/single page support

### LINKED ISSUE #14: Need PDF output

jroper:
Fixes #14

A little something I've been working on in my spare time (I've wanted to do this for ages, since before Paradox existed, I wanted to do it for playdoc).

So, there's actually quite a few features in this. First is single page support, this allows generation of a single page site using `paradoxSingle`:

* A few overidden link renderers and directives have been made to rename all anchors/ref links to `#path/to/page.html~anchor`.
* Header rendering is overridden to offset header depth by the pages depth in the forest, and also to wrap the header content in a span to allow inserting a counter via css (otherwise it messes with the anchor).
* toc directive is disabled, doesn't make sense to have a TOC in the middle of a single page document.
* Images are resolved so they can be included in the single page
* A new template, `single.st`, is used to render the page, it's passed the first page as `page`, and the rest of the `pages` as pages.
* toc renderer (for nav) is overridden to support correct link generation
* `single.st` links to both `page.css`, and `single.css`, so if `page.css` has been overridden, `single.st` will just get that.
* `single.css` adds styling counters to number headings up to 4 levels deep, and adds the necessary numbering to the nav too.

The next feature is a print layout support, which can be compiled by running `paradoxPdfSite`:

* Introduced a `print.st` and `print.css`. Basically, foundations grid layout messed up wkhtmltopdfs rendering, so the print template gets rid of all that, and while it's at it, it drops all navs, headers, footers, etc. They aren't needed.
* There's also a `print-cover.st`, this is for rendering the cover page.

Finally, the PDF support, which can be run by running `paradoxPdf`:

* Runs a [wkhtmltopdf](https://wkhtmltopdf.org/) docker image that I built. Using a docker image has a few advantages, it means it works on all platforms that have docker, but it also avoids problems where wkhtmltopdf running on different platforms can produce different results. Biggest disadvantage - the docker image is 2GB.
* Generates a table of contents. There's a `print-toc.xslt` (xslt!!) that converts the generated overview xml file from wkhtmltopdf to a table of contents, this includes the `page.css`, `single.css` and `print.css` stylesheets, so the TOC is styled by those, and by default, it will number all the headings in the same way that they are numbered in the document.
* The output of `print-cover.st` is used as the cover page.
* Everything seems to work, page numbers, footers, it renders nicely, prettify works.

The regular paradox config for all these commands can be customised, so for example if you want to pass custom properties just for the PDF, use `paradoxProperties in (Compile, paradoxPdf)`, it defaults to the normal one, so you can append to it too to add or override properties.


jroper:
And of course, an example output file: [cloudstate-documentation.pdf](https://github.com/lightbend/paradox/files/3609980/cloudstate-documentation.pdf).

patriknw:
Woot!

jroper:
I added tests, appveyor build failed due to not having docker installed. Not sure if we should fix by adding Docker to appveyor, or by skipping that test on appveyor.

sfxcode:
Hi, is there a chance to release this pull request?

I like the idea of the PDF export ...

raboof:
I'd be OK with skipping this test on appveyor and getting it in. @sfxcode would you like to give that a shot?

sfxcode:
> I'd be OK with skipping this test on appveyor and getting it in. @sfxcode would you like to give that a shot?

I use paradox at work and in my open source projects. I will start testing this feature as soon as possible. Thx for your fast reply and greeting from Wuerzburg, Germany

raboof:
Looking forward to it! Waving from Deventer, the Netherlands :wave:

jroper:
I've updated the PR so that the pdf scripted test doesn't run on appveyor - I did this by moving it from `src/sbt-test/paradox/pdf` to `src/sbt-test/requires-docker/pdf`, and then configuring appveyor to run `scripted paradox/*` as one of its commands.

jroper:
And I've rebased to fix a compile error that was introduced by rebase.

jroper:
Build passes, ready for merge (and release).

PR #376, opened 2019-10-07
Link titles for scaladoc/javadoc

ennru:
* Show the FQCN as link title on source directives (This is helpful to show the reader the package name without following the link.)

* Overwrite a link title with the `title` attribute on `@link`


raboof:
(travis failure is #377)

PR #385, opened 2019-11-19
Javadoc directive: set link style per package

ennru:
Since Javadoc 9 API docs do not use frames anymore and the URLs to class documentation changed.
https://github.com/lightbend/paradox/pull/355 introduced choosing the link style via
Paradox properties and dependent on the host JVM when creating docs.

This adds support to choose the link style per base URL via Paradoc properties
as different API docs use different versions of Javadoc to create it.


tests/src/test/scala/com/lightbend/paradox/markdown/JavadocDirectiveSpec.scala
```diff
@@ -100,4 +100,14 @@ class JavadocDirectiveSpec extends MarkdownBaseSpec {
       html("""<p><a href="http://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/org/reactivestreams/Publisher.html" title="org.reactivestreams.Publisher"><code>Publisher</code></a></p>""")
   }
 
+  it should "support choosing direct links for per package" in {
```
raboof:
```suggestion
  it should "support configuring 'direct links' per package" in {
```

Issue #187, opened 2018-01-26
Warn when linking to a .md file


raboof:
'normal' links take a .html parameter (`[foo](foo.html)`), while `@ref` links and ToC entries expect `[foo](foo.md)`.

This can be easy to mix up: `@ref`s and ToC entries are checked, but `[foo](foo.md)` will just be a broken link.

Should we issue a warning or even throw an error when we encounter a relative non-`@ref` link to a `.md` file?

PR #393, opened 2019-12-02
Fail for regular links to .md

ennru:
Add link validation via a regex pattern to solve #187.

The default pattern allows external links to markdown files, eg. a Github readme page may be linked.


core/src/main/scala/com/lightbend/paradox/markdown/Writer.scala
```diff
@@ -92,6 +93,10 @@ object Writer {
 
   val DefaultSourceSuffix = ".md"
   val DefaultTargetSuffix = ".html"
+  val DefaultIllegalLinkPath =
+    // #DefaultIllegalLinkPath
+    raw"""^(?!https?:).*\.md(#.*)?""".r
```
raboof: Why the `^(?!https?:).*`?

ennru: The problematic links are the local links within the Paradox generated docs. External links may point to eg. Github .md files such as https://github.com/lightbend/config/blob/master/README.md#using-the-library

raboof: aaah I misread the regex :smile: :+1:

PR #394, opened 2019-12-02
Fix default javadoc base for jdk11+

raboof:
Kind of dodgy that we rely on sys.props.get("java.specification.version")
but ok.


ennru:
And there is this other use (which is more about #385, acutally)
https://github.com/lightbend/paradox/blob/7b45d589287e5e9f2a007550cf6db69f2c47fcff/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala#L333

ennru:
The tests fail on JDK 11 now...

PR #403, opened 2020-01-09
Make paradox/default-theme succeed regardless of current year

raboof:


plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala
```diff
@@ -261,7 +261,9 @@ object ParadoxPlugin extends AutoPlugin {
 
   def dateProperties: Map[String, String] = {
     import java.text.SimpleDateFormat
-    val now = new java.util.Date
+    val now = sys.env.get("SOURCE_DATE_EPOCH")
```
ennru: Where does this environment value come from? Isn't the overwrite in the test enough?

raboof: Sorry, that was unclear. SOURCE_DATE_EPOCH is a standard, and at first I figured supporting it would be a good way to solve this issue - but I changed my mind and ended up updating `paradoxProperties` instead. Supporting `SOURCE_DATE_EPOCH` might still be nice and since the change is pretty simple/noncontroversial I left it in there.

Do you want me to move it to a separate PR?

(unfortunately you can't pass environment variables to scripted tests, so that makes it less applicable here)

ennru: Ok, that is what I thought.

Issue #426, opened 2020-04-07
Warn about scala binary version suffix on dependencies


ignasi35:
The [`@@dependency`](https://developer.lightbend.com/docs/paradox/current/directives/dependencies.html) directive renders `sbt`/Maven/Gradle/... snippets to easily copy paste a library dependency on the users build code.

The dependencies usually include a `scala.binary.version` suffix which can lead users unfamiliar with the scala ecosystem to mixing artifacts such as `akka-actor_2.13` and `akka-testkit_2.12` which leads to obscure errors (e.g. `noclassdeffounderror scala/serializable`).  

The `@@dependency` could, instead, not set the `scala.binary.version` to a given value and use a build-specific placeholder (e.g. Maven `property`) with a small comment similar to the support to [symbol/value](https://raw.githubusercontent.com/akka/alpakka-kafka/master/docs/src/main/paradox/testing.md) in:

```
@@dependency [Maven,sbt,Gradle] {
  group=com.typesafe.akka
  artifact=akka-stream-kafka-testkit_$scala.binary.version$
  version=$project.version$
  scope=test
  symbol2=AkkaVersion
  value2="$akka.version$"
  group2=com.typesafe.akka
  artifact2=akka-stream-testkit_$scala.binary.version$
  version2=AkkaVersion
  scope2=test
}
```

[rendering](https://doc.akka.io/docs/alpakka-kafka/current/testing.html#testing) as

```xml
    <properties>
      <akka.version>2.5.23</akka.version>
    </properties>
    <dependency>
      <groupId>com.typesafe.akka</groupId>
      <artifactId>akka-stream-kafka-testkit_2.12</artifactId>
      <version>2.0.2</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>com.typesafe.akka</groupId>
      <artifactId>akka-stream-testkit_2.12</artifactId>
      <version>${akka.version}</version>
      <scope>test</scope>
    </dependency>
```

But such that any paradox user gets the symbol/value usage out fo the box (instead fo copy/pasting it every single time).

This came up during a [gitter discussion](https://gitter.im/akka/akka?at=5e8c822c9316f34b8d870ed3).

So, a paradox user simply adding:

```
@@dependency [Maven,sbt,Gradle] {
  group=com.typesafe.akka
  artifact=akka-stream-kafka-testkit_$scala.binary.version$
  version=$project.version$
  scope=test
}
```

would get the following maven snippet for free:

```xml
    <properties>
[...]

PR #429, opened 2020-04-16
Dependency directive: render symbol for Scala binary version

ennru:
Make the dependency directive detect the use of `scala.binary.version` variable and render it as symbol for Maven and Gradle users.

The intend is to make it more likely for users of these build tools to pick the scala-versioned dependencies with a consistent Scala binary-compatible version.

Refs #426


ignasi35:
Oh, I don't have powers to merge :-(

PR #397, opened 2019-12-03
Scaladoc: better support for inner classes

ennru:
Allow references to inner classes in `@scaladoc` with either a `.` (inline with `@javadox` or `$$` as the separator.

Follow up to #395 
Preparation for https://github.com/lightbend/sbt-paradox-apidoc/issues/86

PR #395, opened 2019-12-02
Linking to javadoc inner classes

raboof:

Issue #98, opened 2017-02-14
Please support Pandoc 


zoosky:
See http://pandoc.org/MANUAL.html
Why? Generate epub

eed3si9n:
Having Pandoc support would also allow PDF output a la Pamflet.

andyczerwonka:
If someone doesn't grab this one sooner, I'll take a look at this one on or around July 1. https://twitter.com/andyczerwonka/status/874814697752494080

andyczerwonka:
I would suggest that this story is not about supporting Pandoc, it's about generating both PDF and ePub for offline reading. Pandoc is an implementation detail - does anyone disagree? Is there some other reason for the Pandoc dependency?

cleverdeveloper:
@andyczerwonka any updates?

andyczerwonka:
@cleverdeveloper I pulled the repository a couple of days ago and started to look at the build process and where to inject the documentation steps. I should have more to share this weekend.

andyczerwonka:
Apologies to all, I won't have an update until I get back from vacation mid-August.

andyczerwonka:
All, shall we close this issues? At the moment this issue is highly dependent on #81 given there is discussion that takes PDF into account.

eed3si9n:
> All, shall we close this issues? At the moment this issue is highly dependent on #81 given there is discussion that takes PDF into account.

I don't see a reason to close this issue, just because there's a related discussion happening elsewhere. We can close it when Paradox is able to generate a Pandoc-compatible single markdown file. In my opinion, it's more of a matter of String concatenation rather than markdown parser implementation.

andyczerwonka:
I think that's fair, but there is clearly a dependency. We may be able to solve the PDF generation depending on library that is chosen, which is really the ask. Pandoc is an implementation concern.

eed3si9n:
ok. I can rename this issue to support EPUB generation.

huntc:
Just wondering if there's been any further thinking on an implementation approach to solve the use case of publishing paradox docs. I might attempt to solve this over the next couple of months for my own purposes.

NightMachinary:
Any updates on epub support?

PR #431, opened 2020-04-21
API doc directives: accept full character range in package names

poWer4aiX:
According to the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se8/html/index.html) member of packages, e.g. subpackages are identifiers. Identifiers itself is

>  ...an unlimited-length sequence of Java letters and Java digits, the first of which must be a Java letter. 

As a `Java letter` includes more than what is expressed by the regex `\b[a-z][a-z0-9_]*)\.`, package names which include uppercase or non ASCII characters cannot be processed. I know that there are some _naming conventions_ flying around but something like `aBc.DE.fg` is still a valid package name in my opinion. 

I have changed the regex in `packageDotsToSlash()` to `(\b\p{javaJavaIdentifierStart}\p{javaJavaIdentifierPart}*)\.`. Unfortunately this breaks with the changes applied for issue #397, #395, #98, #86 to handle inner classes. In order to support this notation, the regex can be relaxed to `(\b\p{javaLowerCase}\p{javaJavaIdentifierPart}*)\.`.

I have added the variable 'scaladoc.strictPackageIdent' (which defaults to `false`) by which you can switch between the two regexp. Once set to true, the package name can conform to an identifier now.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
+++ b/core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
@@ -312,25 +312,53 @@ abstract class ApiDocDirective(name: String)
 }
 
 object ApiDocDirective {
-  /** This relies on the naming convention of packages being all-ascii-lowercase (which is rarely broken), numbers and underscore. */
-  def packageDotsToSlash(s: String): String = s.replaceAll("(\\b[a-z][a-z0-9_]*)\\.", "$1/")
+  /**
+   * Converts package dot notation to a path, separated by '/'
+   * Allow all valid java characters and java numbers to be used, according to the java lang spec.
+   *
+   * @param s package or full qualified class name to be converted.
+   * @param packageNameStyle Setting `startWithLowercase`` will get it wrong when a package name
+   *                         starts with an uppercase letter or when an inner class starts with
+   *                         a lowercase character, while `startWithAnycase` will derive the wrong
+   *                         path whenever an inner class is encountered.
+   * @return Resulting path.
+   */
+  def packageDotsToSlash(s: String, packageNameStyle: String): String =
+    if (packageNameStyle == "startWithAnycase")
+      s.replaceAll("(\\b\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)\\.", "$1/")
+    else
+      s.replaceAll("(\\b\\p{javaLowerCase}\\p{javaJavaIdentifierPart}*)\\.", "$1/")
+}
+
+object ScaladocDirective {
+  final val ScaladocPackageNameStyleProperty = raw"""scaladoc\.(.*)\.package_name_style""".r
+
 }
 
 case class ScaladocDirective(ctx: Writer.Context)
   extends ApiDocDirective("scaladoc") {
 
+  import ScaladocDirective._
+
+  val defaultPackageNameStyle = variables.getOrElse("scaladoc.package_name_style", "startWithLowercase")
+  val packagePackageNameStyle: Map[String, String] = variables.collect {
+    case (property @ ScaladocPackageNameStyleProperty(pkg), url) => (pkg, variables(property))
+  }
+
   protected def resolveApiLink(link: String): Url = {
     val levels = link.split("[.]")
     val packages = (1 to levels.init.length).map(levels.take(_).mkString("."))
-    val baseUrl = packages.reverse.collectFirst(baseUrls).getOrElse(defaultBaseUrl).resolve()
-    url(link, baseUrl)
+    val packagesDeepestFirst = packages.reverse
+    val baseUrl = packagesDeepestFirst.collectFirst(baseUrls).getOrElse(defaultBaseUrl).resolve()
+    val packageNameStyle = packagesDeepestFirst.collectFirst(packagePackageNameStyle).getOrElse(defaultPackageNameStyle)
+    url(link, baseUrl, packageNameStyle)
   }
 
   private def classDotsToDollarDollar(s: String) = s.replaceAll("(\\b[A-Z].+)\\.", "$1\\$\\$")
 
-  private def url(link: String, baseUrl: Url): Url = {
+  private def url(link: String, baseUrl: Url, packageNameStyle: String): Url = {
     val url = Url(link).base
-    val path = classDotsToDollarDollar(ApiDocDirective.packageDotsToSlash(url.getPath)) + ".html"
+    val path = classDotsToDollarDollar(ApiDocDirective.packageDotsToSlash(url.getPath, packageNameStyle)) + ".html"
     (baseUrl / path).withFragment(url.getFragment)
   }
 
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

poWer4aiX:
Stupid question, is there anything I have to do in order to pass the check? I don't understand why "the committer (me) should not linked to a user"...

lightbend-cla-validator:
Hi @poWer4aiX,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

poWer4aiX:
CLA signed and replied by email...

raboof:
Your CLA signature somehow didn't make it into our system yet. That's probably a problem on our side - how did you submit it?

poWer4aiX:
Hi,

second try by using "reply all". Simply replying, resulted i a mail send
to "notifications@github.com", which seems to be wrong.

Thanks,
Christian

Am 21.04.2020 um 20:21 schrieb Lightbend CLA Validator:

poWer4aiX:
@raboof, I simply filled out the PDF and attached it by replying to the email I received. Just noticed that this was addressed to `notifications@github.com` which is obviously the wrong address, I guess. 

Resent it some minutes ago to `reply+...`, hope this is working now. Sorry for any inconveniences.

lightbend-cla-validator:
Hi @poWer4aiX,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

ennru:
Your CLA still hasn't shown up in our system, I'm sorry.
Can you try and sign it with the Github integration?

poWer4aiX:
Can you please explain how to "sign it with the Github integration" or provide me a link with a description? BTW, I just sent it to "cla@lightbend.com".

ennru:
Sorry for all this trouble, on https://www.lightbend.com/contribute/cla you'll find the "Confirm with your GitHub account" link.

ennru:
Ok, all good now. I checked the CLA manually.

poWer4aiX:
Thanks for your support.

Well in the latest version I have applied the change for `@scaladoc` and `@javadoc`. In addition I followed I added a setting for `[javadoc|scaladoc].<package-prefix>.package_name_style`, similar to `jadadoc...link_style`. Defaults to `startWithLowercase`, can be set to `startWithAnycase`.

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -311,25 +311,53 @@ abstract class ApiDocDirective(name: String)
 }
 
 object ApiDocDirective {
-  /** This relies on the naming convention of packages being all-ascii-lowercase (which is rarely broken), numbers and underscore. */
-  def packageDotsToSlash(s: String) = s.replaceAll("(\\b[a-z][a-z0-9_]*)\\.", "$1/")
+  /**
+   * Converts package dot notation to a path, separated by '/'
+   * Allow all valid java characters and java numbers to be used, according to the java lang spec. In order to
+   * be allow to convert `package.Outer.Inner` in scala to the `packge/Outer$$Inner`, `strictPackageIdent` has
+   * be set to `true`.
```
raboof, line 318: Seems like this is no longer accurate. Good to give an example: `startWithLowercase` will get it wrong when a package name starts with an uppercase letter or when an inner class starts with a lowercase character, while `startWithAnycase` will derive the wrong path whenever an inner class is encountered. Does that summarize the trade-off correctly?

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -311,25 +311,53 @@ abstract class ApiDocDirective(name: String)
...
+   * @param s package or full qualified class name to be converted.
+   * @param packageNameStyle  `startWithLowercase` requires subpackages to start with a lowercase character
+   *                          (which is the default to be backwards compatible).
+   *                          `startWithAnycase` allows packages to start with uppercase also.
+   * @return Resulting path.
+   */
+  def packageDotsToSlash(s: String, packageNameStyle: String = "startWithLowercase") =
```
raboof, line 325: This default isn't used anywhere, right?
```suggestion
  def packageDotsToSlash(s: String, packageNameStyle: String) =
```

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -311,25 +311,53 @@ abstract class ApiDocDirective(name: String)
...
+  final val ScaladocPackageNameStyleProperty = raw"""scaladoc\.(.*)\.package_name_style""".r
+
 }
 
 case class ScaladocDirective(ctx: Writer.Context)
   extends ApiDocDirective("scaladoc") {
 
+  import ScaladocDirective._
+
+  val defaultPackageNameStyle = variables.getOrElse("scaladoc.package_name_style", "startWithLowercase")
```
raboof, line 342: :+1: to keep the default

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
poWer4aiX: You're right, it's the old style of setting. Will shorten this block and add your description to the `@param` description.

```suggestion
   * Converts package dot notation to a path, separated by '/'
   * Allow all valid java characters and java numbers to be used, according to the java lang spec.
   *
   * @param s package or full qualified class name to be converted.
   * @param packageNameStyle Setting `startWithLowercase`` will get it wrong when a package name
   *                         starts with an uppercase letter or when an inner class starts with
   *                         a lowercase character, while `startWithAnycase` will derive the wrong
   *                         path whenever an inner class is encountered.
```

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
poWer4aiX: Yes it is not explicitly checked. In the first thought I wanted to define it as simple flag, but using it as a style is more flexile. Was also wondering if it makes sense to add the old style where ascii characters were allowed only, but skipped this. Do you think it is worse to add?

raboof: I don't think we need the old behavior anymore :+1:

poWer4aiX:
If you think the change in general is fine, I will also update the documentation.

poWer4aiX:
@raboof in `@javadoc` the cases with inner classes

 - having a subpackage starting with an uppercase character (org.example.Lib.Outer.Inner)
 - having an outer class starting with a lowercase character (org.example.lib.outer.Inner)
 - having the inner class starting with a lowercase character (org.example.lib.Outer.inner)

are not addressed so far. `@javadoc` will resolve to an incorrect URL in all cases. To fix this I propose to make use of the opposite from scala and resolve `$$` in the FQCN to a `.`.

I have implemented it and added the corresponding tests as well.

core/src/main/scala/com/lightbend/paradox/markdown/Directive.scala
```diff
@@ -371,14 +391,31 @@ case class JavadocDirective(ctx: Writer.Context)
...
+    val packageNameStyle = packagesDeepestFirst.collectFirst(packagePackageNameStyle).getOrElse(defaultPackageNameStyle)
+    url(link, baseUrl, linkStyle, packageNameStyle)
+
+  }
 
+  private def dollarDollarToClassDot(s: String) = s.replaceAll("\\$\\$", ".")
+
+  private[markdown] def url(link: String, baseUrl: Url, linkStyle: LinkStyle, packageNameStyle: String): Url = {
+    val url = Url(link).base
+    val path = dollarDollarToClassDot(ApiDocDirective.packageDotsToSlash(url.getPath, packageNameStyle)) + ".html"
```
raboof, line 414: So any names *actually* containing `$$` are now broken - I suppose that should be exotic enough that this is OK.

poWer4aiX: Yes I also think that this might be a very theoretical case. If we want to get rid of it, we can also think of using a non java character, e.g. a `..` (two dots) instead of `$$`. If I', right then this is invalid for any full qualified name.

mdedetrich:
Is there a plan to merge this PR? This is mainly in relation to https://github.com/lightbend/paradox/issues/515 but in general its been approved but sitting here unmerged for 2 years.

mdedetrich:
Perfect, thank you!

PR #437, opened 2020-05-23
Cross compile `core` with Scala 2.13

joan38:
I am looking into writing a plugin to make sites with Mill and Paradox. My current blocker is that Mill is on 2.13.

Thanks

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,6 @@ import java.lang.management.ManagementFactory
 inThisBuild(List(
   organization := "com.lightbend.paradox",
   licenses += "Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0.html"),
-  scalaVersion := "2.12.10",
   organizationName := "lightbend",
   organizationHomepage := Some(url("https://lightbend.com/")),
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -44,6 +44,7 @@ lazy val core = project
   .in(file("core"))
   .settings(
     name := "paradox",
+    crossScalaVersions := Seq(scalaVersion.value, "2.13.2"),
```
ennru, line 47: Isn't cross-compiling the `core` module enough? This will even cross-compile the sbt plugin and the testkit.

joan38: I thought this would cross compile the core since it's applied on `lazy val core`.
How do you cross compile the core?

ennru: Ah, sorry. You're right. To the lazy me Github showed it looked like it was `inThisBuild`.
I'll have a closer look...

joan38:
@ennru I removed the one in `inThisBuild` since it will take the on in the `AutoPlugin` anyway.

ennru:
Thank you for your interest in brining Paradox to a broader audience!

joan38:
Thanks!

PR #449, opened 2020-08-27
Fix snippet label regex to allow `w`

pvlugter:
Discovered that the snippet label matching was incorrect, when trying examples shared by both paradox and asciidoc — having both paradox labels and asciidoc tags in the same comment worked, apart from when the tag contained a lowercase `w`. Original intention was for this to not be possible, with either the label at the end of the line, or followed only by whitespace then non-word characters.

Added the missing backslash. The unicode shrug in the test can no longer have underscores.


ennru (approved):
LGTM.

jrudolph:
> The unicode shrug in the test can no longer have underscores.

I found the humerus removal quite humorous, paradoxically.

PR #461, opened 2020-11-26
Add SameSite attribute to paradoxGroups cookie

marcospereira:
Not specifying `SameSite` implies the value is `None`, but then the cookie should be marked as `Secure`. Browsers will start to reject cookies that are `SameSite=None` but not `Secure`.

This is the message I'm getting on Telemetry/Akka docs when using Firefox:

Screen Shot 2020-11-26 at 3 43 08 PM

The link at the message: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

### DIFF
--- a/themes/generic/src/main/assets/js/groups.js
+++ b/themes/generic/src/main/assets/js/groups.js
@@ -122,9 +122,8 @@ $(function() {
       .val(group);
 
     // Inline snippets:
-    for (var i = 0; i < catalog[supergroup].length; i++) {
-      var peer = catalog[supergroup][i];
-      if (peer == group) {
+    catalog[supergroup].forEach(peer => {
+      if (peer === group) {
         $("." + group).show();
       } else {
         $("." + peer).hide();
@@ -142,9 +141,7 @@ $(function() {
       });
     });
 
-    for (var i = 0; i < groupChangeListeners.length; i++) {
-      groupChangeListeners[i](group, supergroup, catalog);
-    }
+    groupChangeListeners.forEach(listener => listener(group, supergroup, catalog));
   }
 
   function switchToTab(dt) {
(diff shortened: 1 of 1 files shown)


themes/generic/src/main/assets/js/groups.js
```diff
@@ -19,30 +19,24 @@ $(function() {
...
-    if(!exdays) exdays = 365;
-    var d = new Date();
-    d.setTime(d.getTime() + (exdays*24*60*60*1000));
-    var expires = "expires=" + d.toGMTString();
-    document.cookie = cname + "=" + encodeURIComponent(cvalue) + ";" + expires + ";path=/";
+  // https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
+  function setCookie(cookieName, cookieValue, daysToExpire) {
+    if (!daysToExpire) daysToExpire = 365;
+    const now = new Date();
+    now.setDate(now.getDate() + daysToExpire);
```
marcospereira, line 26: This adds the number of days to the date instead of calculating the number of seconds as it was before.

marcospereira (commented):
We later need to copy the new `groups.js` file to the themes were the same problem happens. I suspect Akka Paradox Theme and the one used by Lightbend Telemetry docs are the ones we need to care about.

themes/generic/src/main/assets/js/groups.js
```diff
@@ -128,9 +122,8 @@ $(function() {
       .val(group);
 
     // Inline snippets:
-    for (var i = 0; i < catalog[supergroup].length; i++) {
-      var peer = catalog[supergroup][i];
-      if (peer == group) {
+    catalog[supergroup].forEach(peer => {
```
raboof, line 125: I'm never sure if such things are available on all browsers we care about :D - :+1: I guess

raboof (approved):

themes/generic/src/main/assets/js/groups.js
marcospereira: Me too. 😅 

But, yeah, it is widely supported according to https://caniuse.com/mdn-javascript_builtins_array_foreach.

PR #462, opened 2020-12-03
Dependency directive: produce BOM notation, modern Gradle

ennru:
- Add support to render Maven and Gradle notation to reference Bill of Materials POMs.
- Update the Gradle notation to recent Gradle (`implementation` instead of `compile` scope, correct notation for the symbolic version map).
- Add the `<dependencies>` markers to the Maven notation to produce more copy-and-paste friendly XML.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: scala
 
 env:
   matrix:
-    - TRAVIS_JDK=adopt@1.8.0-222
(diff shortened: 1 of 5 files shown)


ignasi35 (approved):

ennru:
(I just noticed that the migration to Travis-ci.com is not done on this repo, I don't have powers to commence it.)

PR #474, opened 2021-02-18
Code snippets now contain a button copying the contents

johanandren:
Two text buttons up top right in snippets now: `[ source ]  [ copy ]`

This aligns the snippets with other doc tools that has a copy button up top right.


ennru:
The functionality looks good to me.

The design less so (from Paradox own documentation)
image

The scripted test need to be adapted.

johanandren:
Ah, forgot to check the default theme after getting it to work with the akka one indeed. I'll look into the scripted tests as well.

johanandren:
Better now hopefully

Screenshot 2021-02-18 at 16 25 39

johanandren:
Both chrome and firefox messes up the fonts on those buttons though, making them the code font, I'll see if I can fix.

ennru (approved):
LGTM.

The next request will be to have the "copy" button on all markdown verbatim sections...

PR #490, opened 2021-05-29
Retry count for validate links task

### LINKED ISSUE #488: Support retry count for validate link tasks
We are observing intermittent build failures in our CI due to failure of `paradoxValidateLinks` tasks.

We noticed few of the external links we have in our documentation are very slow and returns some errors for the first time.

It would be nice feature to have, to configure retry counts for validate link tasks.

kpritam:
Fix #488

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
+++ b/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
@@ -503,7 +503,7 @@ object Validator {
   //502 Bad Gateway
   //503 Service Unavailable
   //504 Gateway Timeout
-  private val retryableStatusCodes = List(500, 502, 503, 504)
+  private val retryableStatusCodes = Set(500, 502, 503, 504)
 
   def validateWithRetries(request: Connection, retryCount: Int): Connection.Response =
     try {
--- a/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
+++ b/core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
@@ -19,10 +19,10 @@ package com.lightbend.paradox
 import com.lightbend.paradox.template.PageTemplate
 import com.lightbend.paradox.markdown._
 import com.lightbend.paradox.tree.Tree.{ Forest, Location }
+
 import java.io.{ File, FileOutputStream, OutputStreamWriter }
 import java.nio.charset.StandardCharsets
-
-import org.jsoup.Jsoup
+import org.jsoup.{ Connection, Jsoup }
 import org.jsoup.nodes.Document
 import org.pegdown.ast._
 
@@ -108,6 +108,7 @@ class ParadoxProcessor(reader: Reader = new Reader, writer: Writer = new Writer,
     groups:           Map[String, Seq[String]],
     properties:       Map[String, String],
     ignorePaths:      List[Regex],
+    retryCount:       Int,
     validateAbsolute: Boolean,
     logger:           ParadoxLogger): Int = {
 
@@ -151,7 +152,7 @@ class ParadoxProcessor(reader: Reader = new Reader, writer: Writer = new Writer,
               reportErrorOnSources(errorCollector, c.allSources)(s"Could not find path [${uri.getPath}] in site")
           }
         case absolute if validateAbsolute =>
-          validateExternalLink(absolute, errorCollector, logger)
+          validateExternalLink(absolute, retryCount, errorCollector, logger)
         case _ =>
         // Ignore
       }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
```diff
@@ -160,20 +161,29 @@ class ParadoxProcessor(reader: Reader = new Reader, writer: Writer = new Writer,
...
         .followRedirects(false)
         .ignoreHttpErrors(true)
         .ignoreContentType(true)
         .execute()
 
+      @tailrec
+      def validateWithRetries(retryCount: Int): Connection.Response = {
+        val res = tryConnect
+        if (retryCount == 0 || res.statusCode() == 200) res
+        else validateWithRetries(retryCount - 1)
```
johanandren, line 182: Will it likely be the status codes that signals temporary unavailable rather than not being able to connect for example (so perhaps throws and is caught further down rather than retried right now)? 

Also - should it be more selective with what error codes it retries for (at least some are permanent or likely permanent 400-407 for example), and perhaps back off a little since immediate retry if the server is overloaded will likely fail again.

kpritam: That make sense. I was specifically targeting use case where site is slow and jsoup times out. Was hoping at least one of the retries will succeed. Was planning to keep it simple unless people file issues/enhancements 😉

I ll look into status codes.

Do you think fist class backoff is necessary? I can definitely try simplest backoff solution.
Have not looked at code completely, but does link check happen in parallel? If not, backoff will introduce lot of delay.

johanandren: So if it is mostly about timeouts, would jsoup still return a response with a status code rather than throw an exception?

I don't think backoff is necessary.

kpritam: Yes, you are right, it should throw exception if endpoint does not return any response.
In my case, some sites were returning `503` for very short period of time.

johanandren: What do you think about explicitly checking for 503 (and perhaps 502, 503 as well, maybe 500) + connection timeout throw and retry only for those rather than across all other (less likely to be temporary) status codes.

kpritam: Yes, I am planning to submit that over the weekend. (Maybe earlier If i get the opportunity)

kpritam: I am retrying following status codes: `500`, `502`, `503`, `504` (included additional `504` - Gateway Timeout) and `SocketTimeoutException`

core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
```diff
@@ -503,3 +496,23 @@ class ParadoxProcessor(reader: Reader = new Reader, writer: Writer = new Writer,
...
 
 }
+
+object Validator {
+
+  //500 Internal Server Error
+  //502 Bad Gateway
+  //503 Service Unavailable
+  //504 Gateway Timeout
+  private val retryableStatusCodes = List(500, 502, 503, 504)
```
johanandren, line 506:
```suggestion
  private val retryableStatusCodes = Set(500, 502, 503, 504)
```

docs/src/main/paradox/validation.md
```diff
@@ -53,3 +53,11 @@ paradoxValidationIgnorePaths ++= Seq(
   "/docs/version/(?!latest).*"
 )
 ```
+
+## Retrying links check
+
+`paradoxValidateLinksRetryCount` setting allows retrying link check for the provided number of times in case link returns non success response code i.e., other than 200 status code.
```
johanandren, line 59:
```suggestion
`paradoxValidateLinksRetryCount` setting allows retrying link check for the provided number of times in case linked server does not reply, or returns a possibly temporary failure response code (500, 502, 503 or 504). 
```

johanandren (commented):
Two suggestions, looks good other than those.

core/src/main/scala/com/lightbend/paradox/ParadoxProcessor.scala
kpritam: Fix pushed.

docs/src/main/paradox/validation.md
kpritam: Fix pushed.

johanandren (approved):

PR #512, opened 2022-04-05
Don't override github.external_url if its already defined

mdedetrich:
So I came across a bug from a specific corner case when trying to integrate [github ref links](https://developer.lightbend.com/docs/paradox/current/directives/linking.html#github-directive) into one of my projects. After some debugging I found the state that causes the issue

1. Have `scmInfo` defined (in my case its because we use ghpages to publish the paradox site to github pages)
2. Define your own custom `github.external_url` in `paradoxProperties`. In my case this is done because we have a `main` branch rather than a `master` branch.

When I tried doing this in my project I noticed that no matter how I overrode `github.external_url` the setting would never applied, after looking into Paradox I found the offending code, i.e. https://github.com/lightbend/paradox/blob/master/plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala#L427-L436

Basically if you have `scmInfo` defined then it will always override a custom defined `github.external_url` because the `linkProperties` just ignores `github.external_url` and returns a hardcoded `github.external_url` value that overrides the original.

### DIFF
--- a/docs/src/main/paradox/directives/linking.md
+++ b/docs/src/main/paradox/directives/linking.md
@@ -137,7 +137,9 @@ a tree revision, for example:
 <https://github.com/lightbend/paradox/tree/v0.2.1>.
 
 If the sbt project's `scmInfo` setting is configured and the `browseUrl`
-points to a GitHub project, it is used as the GitHub base URL.
+points to a GitHub project, it is used as the GitHub base URL. Note that this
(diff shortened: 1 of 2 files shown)


johanandren (approved):
LGTM

mdedetrich:
I added one more commit which updates the documentation with the changed behavior.

pvlugter (approved):

PR #514, opened 2022-04-06
Minor syntax improvements

mdedetrich:
This PR applies syntax improvements that are the result of Intellij's default code inspections.

Some additional notes

* Intellij recommends adding type signatures to methods/fields which are public (hence the reason behind a lot of the addition of manual type annotations). I have the suspicion that a lot of these fields should actually be private and not public (likely the reason why the types were not there in the first place). If so let me know
* `Unit` type annotation was added to any method that happened to return `Unit` (i.e. side effecting methods that don't return anything).
* The various `.toSeq` conversions were removed since Intellij marked them as pointless (being the same type). I am quite sure this is not going to cause issues but I would double check here just to make sure nothing odd happens
* the `@tailrec` annotation doesn't actually change anything in the resulting generated bytecode but it will actually confirm that scalac will correctly translation the recursion into an efficient while loop.
* There were some cases where either `.close()` wasn't being called, or it was but the expression wasn't in a `try`/`finally` block which would mean if there is an exception (for some reason) then the handle's wouldn't be released. I have fixed these cases.

### DIFF
--- a/core/src/main/scala/com/lightbend/paradox/ErrorContext.scala
+++ b/core/src/main/scala/com/lightbend/paradox/ErrorContext.scala
@@ -17,11 +17,11 @@
 package com.lightbend.paradox
 
 import java.io.File
-
 import com.lightbend.paradox.markdown.Page
 import org.pegdown.ast.Node
 
 import scala.collection.mutable.ListBuffer
+import scala.io.BufferedSource
 
 trait ErrorContext {
   def apply(msg: String, index: Int): Unit
@@ -46,7 +46,7 @@ trait ErrorContext {
 class ErrorCollector extends ErrorContext {
   private val errors = new ListBuffer[ParadoxError]()
 
-  private def addError(msg: String, page: Option[File], index: Option[Int]) = {
+  private def addError(msg: String, page: Option[File], index: Option[Int]): Unit = {
     errors.append(ParadoxError(msg, page, index))
   }
 
@@ -62,7 +62,7 @@ class ErrorCollector extends ErrorContext {
 
   def errorCount: Int = errors.toList.distinct.size
 
-  def logErrors(log: ParadoxLogger) = {
+  def logErrors(log: ParadoxLogger): Unit = {
     val totalErrors = errors.toList.distinct
     // First log general errors
     totalErrors.foreach {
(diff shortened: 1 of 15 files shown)


core/src/main/scala/com/lightbend/paradox/ErrorContext.scala
```diff
@@ -78,7 +78,13 @@ class ErrorCollector extends ErrorContext {
       .foreach {
         case (page, errors) =>
           // Load contents of the page
-          val lines = scala.io.Source.fromFile(page)("UTF-8").getLines().toList
+          var source: BufferedSource = null
```
mdedetrich, line 81: Intellij handily picked up the fact that the handle from `source` is never actually closed hence the change here. There are nicer ways of doing this i.e. using `Using` but this is only available in Scala 2.13

core/src/main/scala/com/lightbend/paradox/ErrorContext.scala
```diff
@@ -78,7 +78,13 @@ class ErrorCollector extends ErrorContext {
       .foreach {
         case (page, errors) =>
           // Load contents of the page
-          val lines = scala.io.Source.fromFile(page)("UTF-8").getLines().toList
```
mdedetrich, line 81: Intellij handily picked up the fact that the handle from `source` is never actually closed hence the change here. There are nicer ways of doing this i.e. using Using but this is only available in Scala 2.13

mdedetrich:
So the latest test run is failing on `VarsDirectiveSpec` even though it passes fine locally with JDK8 and the previous github actions test run passed without problems (see https://github.com/lightbend/paradox/actions/runs/2101609267).

Maybe rerunning the tests should work? (although not sure why such a test would be flaky)

johanandren (approved):
LGTM, nice little cleanup, thanks!

johanandren:
Oh, I missed the test failure, which is ofc not LGTM :)

mdedetrich:
@johanandren Can you rerun the tests? The tests passed in a previous run and locally, I actually cannot replicate this error (or something really weird is going on)

pvlugter:
Not sure what happened with the tests, rerun is fine.

pvlugter (approved):
👍🏼 nice cleanup

Issue #515, opened 2022-04-06
Migrate from scalariform to scalafmt


mdedetrich:
I noticed that this project is still using scalariform which is no longer maintained (see last commit at https://github.com/scala-ide/scalariform). Instead the community has moved to scalafmt (which newer akka/lightbend projects are using anyways).

The only issues with doing such a change is it creates a huge diff however since there are not any notable pull requests that are open (apart from https://github.com/lightbend/paradox/pull/431 which is approved but not merged yet, also https://github.com/lightbend/paradox/pull/514 but this is not a feature and hence can easily be merged beforehand).

pvlugter:
Good idea. Maybe a scalafmt config that's fairly aligned with what Akka is using.

mdedetrich:
I was thinking of using the one from sbt-paradox-project-info where I already added scalafmt (https://github.com/lightbend/sbt-paradox-project-info/blob/master/.scalafmt.conf). I also added that same configuration in another couple of lightbend projects, are you happy with that one?

Also should I wait for https://github.com/lightbend/paradox/pull/431 or should I go ahead right now?

pvlugter:
Ok, sounds good. Anything that looks reasonable works.

Looks like #431 will need some conflicts resolved already. But probably best to merge that first. It's been sitting there for a while.

mdedetrich:
Cool I have subscribed to that PR so I will get notified when it gets merged

PR #517, opened 2022-04-07
Replace scalariform with scalafmt

mdedetrich:
This PR replaces scalariform with scalafmt as has been done in other lightbend/akka projects, some additional notes:

* Rather than having the code formatted every time after a `compile` is done, instead we a github action is added which checks that the codebase is formatted when a PR opened. The github action uses scalafmt-native which is very fast (typically checks an entire project in ~5-20 seconds and its also done concurrently from the main Scala build). I have opted for this solution because typically speaking  formatters which trigger on compile/git hook's/ etc etc tend to not play nicely with other tools (such as IDE's, editors). Note that this is in line with the other lightbend/akka projects that I have updated to use scalafmt
* The PR is separated into 3 commits, one for removing scalariform, one that adds scalafmt and one that actually applies it. If you want to change the configuration let me know and I can rebase the PR with the update configuration.

Resolves: https://github.com/lightbend/paradox/issues/515


core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
```diff
@@ -30,20 +30,20 @@ object Frontin {
...
+        source.close()
     apply(lines)
   }
 
   def apply(str: String): Frontin =
     str.linesWithSeparators.toList match {
       case Nil => Frontin(Map.empty[String, String], "")
       case x :: xs if separates(x) =>
-        xs span { !separates(_) } match {
+        xs span !separates(_) match {
```
mdedetrich, line 46: So I managed to find an instance where scalafmt changes the meaning of the program and actually causes code not to compile. Will create a ticket upstream.

mdedetrich: Issue created at https://github.com/scalameta/scalafmt/issues/3182

pvlugter: Looks like fix is coming already, but spelling this out fully so it's not broken for now would be good.

tests/src/test/scala/com/lightbend/paradox/markdown/example.scala
```diff
@@ -18,9 +18,9 @@ package com.lightbend.paradox.markdown
 
 //#github-path-link
 object GithubPathLink {
-  //#github-neither-path-link
+  // #github-neither-path-link
```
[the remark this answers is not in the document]
pvlugter: That should be fine

.scalafmt.conf
```diff
@@ -0,0 +1,13 @@
...
+maxColumn = 120
+docstrings.style = Asterisk
+assumeStandardLibraryStripMargin = true
+rewrite.rules = [RedundantBraces, RedundantParens, SortImports, SortModifiers]
+indentOperator.exemptScope = all
+align.tokens = [
+  {code = "%", owner = "Term.ApplyInfix"},
+  {code = "%%", owner = "Term.ApplyInfix"},
+  {code = ":=", owner = "Term.ApplyInfix"},
+  "="
```
pvlugter, line 12: So no alignment for cases?

```
{code = "=>", owner = "Case"}
```

With everything else aligned, may be good to have this too?

pvlugter (commented):
Cool, thanks for organising this.

core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
mdedetrich: So I would recommend one out of 2 options.

1. Format the entire codebase and then remove, fix the above problem and then temporarily remove `RedundantBraces` until the next scalafmt release.
2. Just wait for the next scalafmt release. In the rare cases where scalafmt breaks code it typically doesn't take long for a new release.

.scalafmt.conf
mdedetrich: Can you provide an example of current Scala code and what you expect the formatted to be? As far as I can see `case` is formatted but maybe I am misunderstanding?

core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
pvlugter: I was thinking just write it in a way that's both formatted and not broken. Either of these should be fine:

```scala
xs.span(!separates(_))
```

```scala
xs span { x => !separates(x) }
```

.scalafmt.conf
pvlugter: I thought I was cases that weren't aligned.

core/src/main/scala/com/lightbend/paradox/markdown/Index.scala
```diff
@@ -148,13 +151,13 @@ object Index {
...
-        // only check first children
-        case first :: _ => linkRef(first, level)
-        case _          => None
-      }
+      case other =>
+        other.getChildren.asScala.toList match {
+          // only check first children
+          case first :: _ => linkRef(first, level)
+          case _ => None
+        }
```
pvlugter, line 159: @mdedetrich example of `case` `=>` that was aligned before, but not any longer

core/src/main/scala/com/lightbend/paradox/markdown/LinkCapturer.scala
```diff
@@ -65,8 +63,8 @@ class LinkCapturer {
 
   private val plugins = Writer.defaultPlugins(Writer.defaultDirectives.map(_.andThen {
     case ext: ExternalLinkDirective => new NodeOverridingDirective(ext)
-    case ref: RefDirective          => new NonRenderingDirective(ref)
-    case other                      => other
+    case ref: RefDirective => new NonRenderingDirective(ref)
+    case other => other
```
pvlugter, line 67: Another example of case statements that are no longer `=>` aligned

.scalafmt.conf
pvlugter: Have commented on a couple of examples where this was aligned before.

core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala
mdedetrich: So I realized that `xs span separates` also works and is the most concise/aesthetically pleasing. Changing it now.

mdedetrich:
So I updated `.scalamft.conf` to align `case` fat arrows but now tests are failing because of the previous issue I posted where in some comments the snippets have extra whitespace

mdedetrich:
Okay so I realized what the issue is, the `SnippetDirectiveSpec` is testing the actual content of the source code at `com/lightbend/paradox/markdown/example.scala` which obviously breaks when you apply a different formatter such as scalafmt or even change the configuration.

Due to this I added another commit called `Apply fixes to SnipDirectiveSpec due to format` which applies fixes so the tests pass but it might be wiser to just add `com/lightbend/paradox/markdown/example.scala` to the `project.excludePaths` (see https://scalameta.org/scalafmt/docs/configuration.html#project) which would mean that this file will never get formatted and hence the possibility of subtly breaking the `SnippetDirectiveSpec` if the formatting configuration changes.

Thoughts?

pvlugter:
> Due to this I added another commit called `Apply fixes to SnipDirectiveSpec due to format` which applies fixes so the tests pass

I think that's fine. Projects using paradox will often have formatting like this enabled as well, so having the same here, where some snippet label comments have whitespace added and others don't, seems a good test.

pvlugter (approved):
Thanks! Let's go ahead with this. We can always tweak things as needed later on.

PR #674, opened 2026-02-26
Cross-build to sbt 2 and Scala 3.

anatoliykmetyuk:
This PR cross-compiles the plugin to Scala 3 and sbt 2, making it possible to use from sbt 2 projects. It's a big PR, so here's a summary of all the changes by category to assist with reviewing. Virtually all the changes introduced by the PR are covered by one of the categories below.

### CI/Testing

1. JTidy's static mutable state caused flaky HTML encoding when tests ran in parallel; set `Test / parallelExecution := false` in core and tests to fix.
2. CI runs `verify` instead of `test`: `githubWorkflowBuild` in `build.sbt` now invokes `verify` (Test/compile, Compile/doc, test, scripted, docs/paradox), so scripted tests run in CI. Also CI is running sbt with `--batch` flag to prevent interactive prompts on project loading failures which are not suitable for automated scripts.
### Scala 3 Compatibility
1. JavaConverters was removed in Scala 3; added version-specific `compat/package.scala` that maps to JavaConverters (2.12) or CollectionConverters (2.13/3) so core cross-compiles.
	1. In Scala 2: `java.util.List.asScala` (JavaConverters) returns `scala.collection.mutable.Buffer`, which extends `scala.collection.Seq`. In Scala 3: `java.util.Collection.asScala` (CollectionConverters) returns `scala.collection.Iterable`, which does not extend `scala.collection.Seq`. Therefore we standardize by adding `.toSeq` where APIs are expecting `Seq[String]`.
2. `com.lightbend.paradox.markdown.Url` had a custom `copy(path, query, fragment)` method that conflicted with the case class's synthetic `copy` in Scala 3; renamed to `withComponents` so core cross-compiles.
3. `MarkdownTestkit`: disambiguated `Path` name collision between `java.nio.file.Path` and `com.lightbend.paradox.markdown.Path`. Scala 2 and 3 use different resolution strategies so a collision happens in Scala 3. The port added explicit `java.nio.file.Path` and `ParadoxPath` alias to avoid relying on context-dependent resolution.
4. Scala 3 requires explicit types for implicit values. Test specs (e.g. `LinkDirectiveSpec`) that use `implicit val context` for the writer context now have an explicit type: `Location[Page] => Writer.Context`.

### sbt 2 Compatibility

1. Wrapped tasks that do not support sbt 2 caching out of the box in `Def.uncached`.
2. Replaced sbt 1 `in` syntax (e.g. `key in scope`) with sbt 2 slash syntax (`scope / key`).
3. Bumped `sbt-web` from `1.5.8` to `1.6.0-M2` to support sbt 2; the new API requires `fileConverter.value` for `syncMappings` and `deduplicateMappings`, due to changes in file types between sbt 1 and sbt 2.
4. [PluginCompat](https://www.scala-sbt.org/2.x/docs/en/changes/migrating-from-sbt-1.x.html#the-plugincompat-technique) and [sbt2-compat](https://github.com/sbt/sbt2-compat) pattern is used for breaking API changes between sbt 1 and sbt 2.
6. sbt 2 changed `ScmInfo.browseUrl` from `java.net.URL` to `java.net.URI` - standardize on `java.net.URI`.
7. `ParadoxPlugin.scala`: inlined `sourcesFor` at its single call site in. Originally it was in the compat layer for sbt 0.13 support, but the plugin no longer cross-builds to sbt 0.13.
8. On sbt 2, `target.value` inside `inConfig(Compile)` resolves to a versioned subdirectory (`target/out/jvm/scala-3.8.1/<project>/`) instead of `target/` as on sbt 1. This makes the output path of the plugin inconsistent. Changed paths to preserve the sbt 1 behavior throughout `ParadoxPlugin.scala`.

### Scripted Tests

1. sbt 2's config parser does not allow hyphens in config names; the `paradox/docs-overlay` scripted test was updated to use camelCase configs (`docsFirst`, `docsSecond`).
2. `paradox/snippet-noindent-writer` and `paradox/snippets` expected HTML updated to match current source files. Scripted tests were not tested by the CI so the html files for those tests were broken by an scalafmt run a while ago.
3. `paradox/libraryDependencies` test was filtered for sbt 2 as the material plugin is not yet ported to sbt 2. Exclusion logic in `build.sbt`: `verify-no-docker` sbt command was changed from a simple alias to a command to enable filtering logic. scripted tests plugin checks for sbt version in build.properties to determine if it matches the sbt version the test is tested against - however this feature does not work in case of glob references, e.g. `paradox/*`. Logic in `verify` command works properly because no globs are used.


johanandren (commented):
I'm not deeploy familiar with all the required changes for sbt-2, but I think this looks good.

anatoliykmetyuk:
@johanandren thank you for taking a look! I've fixed the formatting with a `scalafmt` run, so the CI should pass now.

eed3si9n:
Thanks @anatoliykmetyuk! I can review at sbt 2.x specific parts, but at a glance, the lion's share of the changes seems to be to bring up old code base into Scala 3.x, which includes using 2.13 stdlib.

plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala
```diff
@@ -119,104 +132,114 @@ object ParadoxPlugin extends AutoPlugin {
...
-      (mappings in paradoxTheme).value,
-      (target in paradoxTheme).value
+    paradoxTheme / WebKeys.deduplicators += Def.uncached(SbtWeb.selectFileFrom((paradoxTheme / sourceDirectory).value)),
+    paradoxTheme / mappings := SbtWeb
+      .deduplicateMappings(
+        (paradoxTheme / mappings).value,
+        (paradoxTheme / WebKeys.deduplicators).value,
+        fileConverter.value
+      ),
+    paradoxTheme / target := baseDirectory.value / "target" / "paradox" / "theme" / configTarget(configuration.value),
```
eed3si9n, line 204: Could we use

```scala
target.value / "paradox" / "theme" / configTarget(configuration.value),
```

like it was before, so it's safe against cross building?

eed3si9n (approved):
I have a minor question about target directory, but overall lgtm

plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala
anatoliykmetyuk: The paths seem to differ between sbt 1 and 2 which causes inconsistent behavior. I've mentioned this one at the PR changes above:

> On sbt 2, target.value inside inConfig(Compile) resolves to a versioned subdirectory (`target/out/jvm/scala-3.8.1/<project>/`) instead of `target/` as on sbt 1. This makes the output path of the plugin inconsistent. Changed paths to preserve the sbt 1 behavior throughout ParadoxPlugin.scala.

The current setup should produce consistent paths during cross-build.

anatoliykmetyuk:
The formatting failure seems to be due to the new `using` syntax in the Scala 3 specific sources that scalafmt recognized as invalid. I've set the dialect to `scala3` with the latest commit which should solve it.

johanandren (approved):

